PR #228, opened 2019-04-26
Enh/barrier all 225

### LINKED ISSUE #225: support for barrier_all() operation
currenty barrier() expects a list of qubits. For surface-7, when we have to add a barrier for all the 7 qubits, we have to mention all the 7-qubits. On the same lines, for surface-17, all the 17 qubits would be required as an argument. 

A simple solution to make a barrier for all the qubits while being surface agnostic would be add the support for barrier_all().

imranashraf:
closes #225. Now if no qubit are specified, barrier is applied on all qubits. The API has been updated and **test_barrier.py**  tests the possible variations.

For example the following are equivalent syntaxes:
k.barrier()
k.gate('barrier', [])
k.gate('wait', [], 0)


imranashraf:
@MiguelSMoreira give it a try and let me know so that i can merge this PR.

MiguelSMoreira:
@AdriaanRol , can you take a moment to see if the changes satisfy the initial requirements for the barrier on all qubits? Since this request didn't originate with me, I believe you are better suited to determine whether or not this satisfies our needs...

AdriaanRol:
This looks good to me. One question though, I assume that the wait 0 clocks instruction is equivalent to the barrier instruction. 
If this is (or is not) the case, I think it is good to explicitly include this in the docstring for both the barrier and the wait instruction.

imranashraf:
@AdriaanRol updated the doc strings as per your suggestion.

PR #242, opened 2019-07-17
Link with a correct Python version.

slavoutich:
The problem is coming, if the higher version of Python is present in the system, than one OpenQL is being installed into. CMake is trying to guess the libpython to link to, and it chooses "the best" on its opinion, which is highest, and not one in the environment. Then OpenQL can turn unrunnable, if the required library is not in dynamic library search path.

This happened to us on Windows: CMake found Anaconda and linked against it, despite the fact we don't use Anaconda environment for lab setups now, and this caused a failure, because we used lower version of Python than Anaconda's one.


imranashraf:
thanks @slavoutich . 
@MiguelSMoreira this should solve your problem.

MiguelSMoreira:
Patch tested to work on a measurements computer.
Thank you for your help with this @slavoutich

Issue #260, opened 2019-11-14
Name change required for pypi


AdriaanRol:
There is a naming conflict with an existing package on PyPi. As such we will need to make a name change. 

Carmina will come with a proposal on the name change and how to execute this.

AdriaanRol:
@carminaga

imranashraf:
As per https://github.com/QE-Lab/OpenQL/pull/276, openql is now available on pypi with the distribution name **qutechtools**.

PR #276, opened 2019-11-24
pypi wheel upload

imranashraf:
openql wheels are now available on PYPI. openql can now be installed by

`pip install qutechtools`

updated sphinx documentation

closes issue https://github.com/QE-Lab/OpenQL/issues/260

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ env
 *.cache
 swig/openql.py
 swig/openql/openql.py
+swig/qutechtools.egg-info
 
 # Test files
 test_output
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,7 @@ env
 *.cache
 swig/openql.py
 swig/openql/openql.py
-swig/qutechtools.egg-info
+swig/qutechopenql.egg-info
 
 # Test files
 test_output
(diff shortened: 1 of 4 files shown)


AdriaanRol:
@imranashraf , thanks for putting openql on pypi. 
I know you already updated it but would it be possible to use a different name? There are many tools from qutech and most people in our group know this as openql. Would it be possible to use e.g., qutechtools_openql or something similar? 

I don't really care about the specific name but I think this one will cause a lot of confusion in the future.

imranashraf:
Yes @AdriaanRol , that is possible. **qutechtools_openql** is fine too, but based on [PEP8](https://www.python.org/dev/peps/pep-0008/#package-and-module-names) guidelines, I would suggest using **qutechopenql** to make it shorter and avoid dashes and underscores. What do you think?

imranashraf:
updated to **qutechopenql**

AdriaanRol:
@imranashraf , given that I can now download "qutechopenql" from pypi, is there a reason this PR has not been merged?

imranashraf:
@AdriaanRol i did not know that you can download it, i was waiting for this confirmation and approval. now its merged.

Issue #262, opened 2019-11-14
Better way of describing conditional gate execution


AdriaanRol:
OpenQL is currently a combination of API calls. Expressing classical construct is very counter intuitive. This can be improved.

wvlothuizen:
Below is a quote of an old post of mine on slack channel '#openql-and-friends' that pertains to this subject:

> As some of you know, I'm working on an OpenQL backend for the new Central Controller. This CC differs from the CClight in that it has no notion of quantum operations and internally has a distributed architecture (with a separate processor per connected instrument). In order to be able to support conditional instructions, we need to be able to relate measurements (currently performed on an UHFQC) to actions (performed on VSM or maybe AWG). Because we have no a priori knowledge of the mapping of qubits on either, we devised an end user programmable (i.e. through the backend) Condition Engine.
> This Condition Engine can do many more things than the implicit conditional gates of the CClight (e.g. support expressions on 2 qubits, and support a concept of state), but these are currently difficult/impossible to express in OpenQL.
> Of course we can (and probably will) support the current conditional gates, but I think it would be a great improvement if the condition could be passed as a parameter (preferably a logic expression).
> We also support program flow conditional on measurement results ("comprehensive feedback"). This can only be expressed by the user in 'classical QASM', which is a bit awkward to support for us because it is very close to the CClight internal instruction set, but much less so to ours. As an example, we support a 'loop' instruction which combines decrementing a register and a conditional branch. As another example, in our instructions measurement results live in a register space different from the classic processor registers (namely in distributed shared memory controlled by the real time part of the processor). I think all of those translations would be easier if the classical part of QASM would be a higher level language.
> All in all I think a discussion on classical logic would be appropriate (maybe that is already going on for libqasm2.0, but I'm not connected to that)

wvlothuizen:
meeting homework:

documentation on the CC can be found in the Central Controller dropbox. If you require access please let me know. Places to have a look are "doc/design/CC", "doc/presentations" and "doc/requirements/CC".

wvlothuizen:
meeting homework:

indeed the instruction set of the CC does not operate at the qubit and quantum gate level like the CC-light, but instead it deals with digital inputs (from a UHFQC) and outputs (to an AWG). Moreover, the architecture is distributed: one sequence processor controls one instrument only, and sequence processors can exchange data (measurements) using Distributed Shared Memory.
[...]

PR #283, opened 2019-12-12
Doc/scheduler ccl

jvansomeren:
OpenQL documentation: global picture, top-down, passes, code examples, scheduling and mapping.
But not APIs, installation, rehosting.


jvansomeren:
Just pushed the documentation for review in the meeting of next Thursday, January 9.
Since last time I added:

- scheduling documentation including relation to resources and configuration file entries.
- classical gates documentation, the sequential ones
- control flow support documentation using kernels
- QASM documentation, how it is generated from the internal representation

The compiler passes documentation is work-in-progress; sorry for the repetition there.
All API definitions are missing; I'm leaving them to Imran to add in an automated way.

docs/compiler_passes.rst
```diff
@@ -1,11 +1,181 @@
...
+	of the form ``sqf q0`` to be executed in the same cycle;
+	this is done only when the ``cz_mode`` option has the value ``auto``;
+	such a gate is generated for each operand and for all qubits that need to be detuned;
+	see the detuned_qubits resource description in the CC-Light platform configuration file for details.
+	See :ref:`decomposition`.
+
+- opcode and control store file generation
+	currently disabled as not used by CC-Light
+
+- QISA generation
```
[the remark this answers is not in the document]
jvansomeren: Defined 'platform dependent' pass to mean that its source code is platform dependent.
Such passes are clearly marked now. These are to be described with the platform. The list of passes at the start of the section is just an overview  before the actual descriptions start.

docs/kernel.rst
```diff
@@ -1,4 +1,243 @@
...
+   p.add_for(kloopbody, 10)
+   p.add_kernel(kafterloop)
+
+   # do Dowhileloopbody while (rs1 < rs2)
+   # Afterdowhile
+   kdowhileloopbody = ql.Kernel('Dowhileloopbody', platform, num_qubits, num_cregs)
+   kdowhileloopbody.gate('x', [0])
+   kafterdowhile = ql.Kernel('Afterdowhile', platform, num_qubits, num_cregs)
+   kafterdowhile.gate('y', [0])
+   p.add_do_while(kdowhileloopbody, ql.Operation(rs1, '<', rs2))
```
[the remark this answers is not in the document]
jvansomeren: Added assignments.

docs/platform.rst
```diff
@@ -1,22 +1,326 @@
...
+
+.. code-block:: html
+  :linenos:
+
+  "hardware_settings":
+  {
+      "qubit_number": 7,
+      "cycle_time" : 20,
+
+      "mw_mw_buffer": 0,
```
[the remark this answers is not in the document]
jvansomeren: The rcscheduler (schedule_post179 function) calls insert_buffer_delays as one of its last actions. It is a facility that once was used and doesn't complicate code. Proposal could be to make specification in config file optional.

docs/platform.rst
```diff
@@ -1,22 +1,326 @@
...
+Specifies the list of primitive gates supported by the platform.
+Creation of a primitive custom gate takes its parameters from this specification to initialize the gate's attributes.
+
+Examples of a 1-qubit and a 2-qubit instruction are shown below:
+
+.. code-block:: html
+    :linenos:
+
+	"instructions": {
+		"x q0": {
```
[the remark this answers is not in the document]
jvansomeren: Added the other one as well and added references to definitions of the two forms and to where these are used.

docs/classical_instructions.rst
```diff
@@ -1,4 +1,241 @@
...
+``operation type`` indicates the type of operation which is mainly used for checking
+
+``inv operation`` represents the inverse of the operation; it is used in code generation of conditional branching; see :ref:`kernel`
+
+
+Classical gates in circuits and bundles in the internal representation
+----------------------------------------------------------------------
+
+In circuits and bundles, no difference is made between classical and quantum gates.
+Classical gates are scheduled based on their operands and duration.
```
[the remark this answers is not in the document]
jvansomeren: Yes, they can be in parallel and therefore are scheduled together. What is your point?

docs/kernel.rst
```diff
@@ -1,4 +1,243 @@
+.. _kernel:
+
 Kernel
 ======
 
-contains Kernel details
\ No newline at end of file
+A kernel models a circuit with quantum gates ending in one or more measurements.
```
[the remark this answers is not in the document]
jvansomeren: No, there is not. Updated the text to reflect this.

docs/kernel.rst
```diff
@@ -1,4 +1,243 @@
...
+In OpenQL, this has been extended with:
+
+- control flow that can jump at the end of a kernel to the start of another kernel; see this section
+- classical gates mixed with quantum gates in a single circuit with the design objective to support control flow changes; see :ref:`classical_instructions`
+
+In OpenQL a ``kernel`` is an object; it has a name, a type,
+and a circuit as its main structural attributes.
```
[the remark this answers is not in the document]
jvansomeren: Yes, that is true. That is because control flow cannot be scheduled now. If it could, we could schedule the whole program, with gates having cycles relative to the start of each kernel. There is a note on this in the section control_flow_in_the_internal_representation

docs/classical_instructions.rst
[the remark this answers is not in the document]
wvlothuizen: I guess I don't understand what scheduling of classical instructions really means. For both the CC-light and the CC, there is a single instruction memory with classical instructions and 'quantum' instructions interspersed. 
The quantum instructions are timed, and run in a different pipeline. Normally, the classical instructions run in parallel to the quantum instructions, and thus take no time. The CC even enforces that, it is an error for the 'quantum' pipeline to run empty, because in that case timing would be affected.
Apart from that, at least in the CC classical instructions run at a different rate (5 ns vs. 20 ns), and the duration is not fixed (i.e. due to register dependencies), so making accurate timing predictions is difficult.
Having said this: the CC backend currently does not support classical instructions, and in my opinion the whole OpenQL view op classical instructions should be revised (see issue #262)

imranashraf:
It is very important to split the documentation into User documentation and Developer documentation. IMHO the focus should be on the user documentation in the start so that the user of OpenQL can use it. However, in this case your focus is on Developer's documentation, explaining the internals which a user will not see or touch, hence making it not very useful or interesting for the user.

secondly, it is important to add examples with the explanation. For instance, if a user selects a certain option for scheduling or decomposition, what effect will it have on the generated code !!!

Third, it is generally a good idea to document stuff in separate PRs so that it is easier to review.

jvansomeren:
This documentation was written to address the questions that @AdriaanRol and @MiguelSMoreira have on the use of OpenQL as a compiler and the interplay with the configuration file. It is user documentation, all details in the documentation are needed to understand OpenQL's passes, the resource specifications, and the scheduler. Because the passes don't operate on cQASM but on an internal representation, a description of that one was needed too.
@Imran is still the first author, so feel free to add examples.
But the first priority is to make it available to everyone, by reviewing it and merging it in develop.

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+A circuit defines the operation of a kernel.
+And a program consists of multiple kernels.
+
+Gates can be subdivided in several kinds.
+This is useful in the description of the passes below.
+
+
+First, gates can be subdivided according to where their execution has effect:
+
+- quantum gates; these gates execute in the quantum computing hardware; these gates have at least one qubit as (implicit or explicit) operand; these gates can have classical registers as operand as well and may rely on some execution capability in classical hardware
```
AdriaanRol: Does this distinction relate to classes in the code or is it only a convenient abstraction?

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+
+Quantum gates can be subdivided in several ways from the number of operands they take; this becomes relevant when gates are mapped on the quantum computing platform:
+
+- one-qubit gates; quantum gates operating on one qubit
+
+- two-qubit gates; quantum gates operating on two qubits;
+  these are the main objective in mapping since two-qubit gates require their qubit operands to be connected in the hardware
+
+- multi-qubit gates; quantum gates operating (implicitly or explicitly) on more than two qubits;
+  these must be decomposed to one-qubit and two-qubit gates because more-qubit primitive gates are not supported by the quantum platform
```
AdriaanRol: "because more-qubit primitive gates are not supported by the quantum platform" 
Does this mean that OpenQL is fundamentally incompatible with hardware platforms that specify multi-qubit primitives? 

We have a multiplexed readout operation that is fundamentally multi-qubit.

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+  examples are identity, x, rx(pi), cnot, swap, and toffoli.
+
+- measurement gates; gates that measure out the operand qubits, leaving them in a base state; the measurement result can be left in a classical register
+
+- scheduling gates; gates that only influence execution timing regarding the operand qubits;
+  they provide a cycle window for the qubit state to be operated upon before further use;
+  examples are the wait and barrier gates
+
+
+Particular classes of quantum gates can be further recognized:
```
AdriaanRol: I am assuming that this subdivision has not consequence in the code as in there are no actual `classes` that link to this. 
Pauli gates (and some rotations) are also Cliffords. As such I am confused with how I should interpret this.

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+  (in-line with the purpose of a gate being specialized)
+
+- parameterized gates; custom gates that are not specialized,
+  i.e. with a definition that is not specific for particular qubit operands;
+  all gates created (usually for different qubits) from the same parameterized gate in the platform configuration file
+  have the same semantic attributes
+
+
+
+.. _quantum_gate_attributes_in_the_internal_representation:
```
AdriaanRol: It would be great to link these things to the API autodoc of PR #286 . But probably more something for a later point in time.

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+
+- gates are most directly distinguished by their name
+
+:Note: Distinguishing gates internally in the compiler by their name is problematic; distinguishing by their type (see the table below) would be better; the latter conveys the semantic definition and is independent of the representation (e.g. ``mrx90``, ``mx90``, and ``Rmx90`` all could be names of a -90 degrees X rotation); furthermore, a name is something of the external representation and is mapped to the internal representation using the platform configuration file; however, the enumeration type of type can never include all possible gates (e.g. those with arbitrary angles, any number of operands, etc.) so the type inevitably can be imprecise; but it can be precise when the type refers to the operation only, i.e. excluding the operands
+
+- qubit and classical operands are represented by unsigned valued indices starting from 0 in their respective register spaces
+
+- ``angle`` is in radians; it specifies the value of the arbitrary angle of those operations that need one; it is initialized only from an explicit specification as parameter value of a gate creation API
```
AdriaanRol: I understand why this is done, but this makes it a bit tricky to specify common angles like 180 or 90 degrees (given that there is no definition of Pi available) .

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+
+- ``duration`` is in nanoseconds, just as the timing specifications in the platform configuration file; scheduling-like passes divide it (rounding up) by the cycle_time to compute the number of cycles that an operation takes; it is initialized implicitly when the gate is a default gate or a custom gate, or explicitly from a parameter value of a gate creation API
```
AdriaanRol: Seems like an assumption driven by the hardware backends. I can imagine this not being future proof. 
I would suggest to always do this in seconds and use 10e-9 if one wants to specify nanoseconds (does change ints to floats).

docs/quantum_gate.rst
```diff
@@ -1,4 +1,366 @@
...
+
+#. the qubit and/or classical register operand indices are checked for validity,
+   i.e. to be in the range of 0 to the number specified in the program creation API minus 1
+
+#. if the configuration file contains a definition for a specialized composite gate matching it, it is taken;
+   the qubit parameter substitution in the gates of the decomposition specification is done;
+   each resulting gate must be available as (specialized or parameterized, and non-composite) custom gate,
+   or as a default gate; the decomposition is applied and all resulting gates are created and added to the circuit
+
+#. otherwise, if a parameterized composite gate is available, take it;
```
AdriaanRol: Something went wrong with formatting here. the # is interpreted as a header. Check the rendered docs to see how it looks.

docs/program.rst
```diff
@@ -1,4 +1,41 @@
...
+In the OpenQL programming model, one first creates the platform object and then with it the program object.
+After that, one creates kernels with gates and adds these kernels to the program.
+Finally, one compiles the program and executes it.
+At any time, options can be set and got.
+
+Platform creation takes a name (to use in information messages) and the name of the platform configuration file.
+The latter is used to initialize the platform attributes, e.g. to create custom gates.
+
+A program is created by specifying a name, the platform, and the numbers of quantum and classical registers.
+The latter defaults to 0.
```
AdriaanRol: `0` as an integer, as a string, as a `None` value?  
What does this default mean? 
Might be useful to link to the API autodoc as that is usally more clear than explaining that here.

docs/quantum_gate.rst
jvansomeren: Is a convenient abstraction in the documentation.

docs/quantum_gate.rst
jvansomeren: OpenQL can handle gates with any number of qubits.
See measure, barrier, wait, etc.

Updated the documentation to make clear that the examples given reflect the current mapping pass that targets CC-Light

docs/quantum_gate.rst
jvansomeren: The definitions help in the other parts of the documentation.
Yes, by Cliffords I meant those single-qubit rotations that form a group with 24 equivalence classes. There is an API to create one from this group. And now there is an optimizer that reduces sequences of these to the smallest sequence.

docs/quantum_gate.rst
jvansomeren: Yes, but these are attributes of a gate in the internal representation,
and not necessarily names of parameters in a gate creating API.
I.e. we are free to rename the attributes systematically without affecting the API.
Having given them the same name now seems to lead to confusion.

docs/quantum_gate.rst
jvansomeren: Yes, this has been a choice.
We could as well have defined an attribute pi_fraction instead,
so that the angle would have become pi_fraction*pi
where pi would have been built-in into OpenQL.

docs/quantum_gate.rst
jvansomeren: Is not an issue now.
Will keep this proposal in mind.

docs/program.rst
jvansomeren: I deleted the remark that the number of classical registers has a default 0 when creating a program. Should be in the API definition indeed.

jvansomeren:
Since @AdriaanRol and @MiguelSMoreira and @wvlothuizen have reviewed this pull request,
I propose it is merged now and we close doc/scheduler-ccl.
Further updates can then be done as new issues / extensions on the develop branch.

PR #293, opened 2020-02-26
Mapper

jvansomeren:
Three classes of issues still remain:

1. the documentation in docs still documents develop as it is now, and not the update;
    I'll work on this; since the mapper functionality itself is by default off,
    highest priority has the documentation of the interfaces outside the mapper functionality

2. when running python -m pytest, there are 12 errors;
   executing the tests individually, there are just 2.
   The problem seems to be that option settings from one test carry over to next tests.
   I propose that a new interface is added to src/options to (re)set options to default values
   and to call this interface at the start of each test, before setting the test specific options.

3. some tests don't pass when running them individually (as in 'python test_bugs.py'):
   - test_bugs.py: output files are not equal; there is a problem in using gates of class classical_cc;
     I'll work on this
   - test_parallel_triggers.py: uses qumis_compiler;
     I propose to discard support for it

   test_standard_single_Q_experiments_CBox.py and test_grover.py give no OK nor FAILURE,
   so I assume these are passing

Please note that the external interfaces are not modified. All new stuff is protected by options
that are by default off. So, the update is backward compatible to 'develop'.
Some internal interfaces have changed while adding the mapper functionality.
These changes are inline with what we plan to do for 'modularity'.


wvlothuizen:
please merge 'develop' again into this branch: now the recent documentation changes show up as differences

jvansomeren:
Documentation reflects new external and internal interfaces.
Documentation of mapper functionality still missing.
Bug in tests/test_bugs.py solved.

jvansomeren:
Would be nice when Jenkins would work!
Are there other tests that fail?
I am confused by those red crosses.
I am in car to holiday so I cannot investigate

src/gate.h
```diff
@@ -1549,13 +1555,13 @@ class custom_gate : public gate
             // operation_type = instr["type"];
             l_attr = "duration";
             duration = instr["duration"];
-#if OPT_USED_HARDWARE
```
wvlothuizen: this define is to identify spots where "used_hardware" is referred to. Leaving it in for now makes it easier to remove this unused code later on

src/program.h
```diff
@@ -22,48 +22,53 @@ namespace ql
  */
 class quantum_program
 {
-    protected:
-        bool                        default_config;
-        std::string                 config_file_name;
-        std::vector<quantum_kernel> kernels;
+   protected:
```
wvlothuizen: it looks as if the tab settings are different from rest of project (i.e. 4 spaces)

src/qsoverlay.h
```diff
@@ -0,0 +1,179 @@
...
+#include <gate.h>
+
+
+//Only support for DiCarlo setup atm
+void write_qsoverlay_program( std::string prog_name, size_t num_qubits,
+        std::vector<ql::quantum_kernel>& kernels, const ql::quantum_platform & platform, std::string suffix, size_t ns_per_cycle, bool compiled)
+    {
+
+		//TODO remove the next line. Using this because qsoverlay has some bugs when time is explicited
+		compiled = false;
```
wvlothuizen: messy indentation, also below

src/gate.h
imranashraf: moved back

src/program.h
imranashraf: fixed

src/qsoverlay.h
imranashraf: fixed

src/metrics.h
```diff
@@ -0,0 +1,775 @@
...
+	IOUT(output.str());
+}
+
+namespace ql
+{
+// namespace metrics
+// {
+
+
+// ql::circuit random_circuit_generator(double Nqubits, size_t Ncycles, double idle_fraction, double two_qb_gate_fraction ){
```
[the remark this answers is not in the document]
jvansomeren: The scheduler diff is big because the pre179 code was inadvertently taken out
No harm because post179 is compatible and has proven itself
I propose to leave it like this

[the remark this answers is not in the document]
imranashraf: cleaned some comments, leaving others as it seems work in progress

imranashraf:
all tests pass individually. issues with option and mapper documentation can be dealt separately later. merging for now. for future it will be really easier to review PRs if we make small changes on feature branches and then merge.

PR #295, opened 2020-03-10
Option MA should be AM

QFer:
* Option was wrong for enabling toffoli AM decompositions


imranashraf:
@QFer I would suggest to change the name of the corresponding functions as well in kernel.h to avoid confusion

QFer:
@imranashraf That was why I changed this. In kernel.h (decompose_toffoli) the option is read and checked against "AM" and then the function toff_kernel.controlled_cnot_AM is called. Because the option enabled was "MA" this decomposition was never called. So the bug was in the wrong option name "MA" instead of "AM".

imranashraf:
true. merging

PR #311, opened 2020-04-21
Added documentation for unitary decomposition

anneriet:
I have no write access to the QE-Lab/OpenQL repository, so I cannot directly add this code to that branch, so I added the documentation on my github.

### DIFF
--- a/docs/decomposition.rst
+++ b/docs/decomposition.rst
@@ -38,7 +38,7 @@ TBD
 
 Unitary decomposition
 ^^^^^^^^^^^^^^^^^^^^^
-Unitary decomposion allows a developer of quantum algorithms to specify a quantum gate as a unitary matrix, which is then split into a circuit consisting of ``ry``, ``rz`` and ``cnot`` gates. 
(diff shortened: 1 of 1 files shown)


jvansomeren:
I quickly read the documentation and have some questions:

As far as I remember, @anneriet  wrote your thesis about this and I assume that contains more information on e.g. the algorithm that you employ and exact constraints on the input matrices (any kind of complex values as long as the matrix is unitary?); can you add more on this?

This could e.g. answer the following:
When arbitrary unitary matrices can be input, how exact is the result then?
Which primitive gates to you assume to be supported when generating the circuit?

anneriet:
I added a link to my thesis for the more detailed answers, and the qasm generated to hopefully provide more clarity. The rest of the documentation does not go into a lot of detail, so I was unsure what level to go into here. 
To answer your questions anyway, the result is as exact as the input or as exact as the output, limited by the number of decimals of qasm, the internal computation has the mathematical precision of doubles (about 10^-16). 
And the decomposition generates only three types of gates, this is included in the documentation, so no assumptions are made, it always uses those. To execute the circuit those three need to be supported, or translations from the ry, rz and cnot gates to supported gates, but that is no concern of the decomposition part.

PR #321, opened 2020-05-28
Upstream DCL additions

wvlothuizen:
this is to upstream recent additions in the OpenQL setup in DiCarloLab pertaining to Quantum Inspire support


wvlothuizen:
I'll first fix the build, some conflicts weren't resolved yet, will get back to you when that's done.

I'm aware of the discussion concerning wait/barrier en think we need something better, but then again we really needed to have something working for the Quantum Inspire release. All those changes are inside #define OPT_DECOMPOSE_WAIT_BARRIER, and thus easy to disable.

src/version.h
```diff
@@ -4,4 +4,4 @@
 
 	 NB: OPENQL_VERSION_STRING is also decoded by setup.py
 */
-#define OPENQL_VERSION_STRING "0.8.1.dcl-3"
+#define OPENQL_VERSION_STRING "0.8.1.dcl3"
```
razvnane, line 7: This is not a legal version for the module ("dcl" is not a recognized string)  and most probably what causes the CI to fail with ImportError. Legal names should follow this rule

[N!]N(.N)*[{a|b|rc}N][.postN][.devN]

according to https://www.python.org/dev/peps/pep-0440/

This is confirmed by compiling this PR locally and getting this warning:
<path>/python3.7/site-packages/setuptools/dist.py:484: UserWarning: The version specified ('0.8.1.dcl3') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version

src/version.h
```diff
@@ -1,7 +1,8 @@
 /* Version examples:
 	"0.8.0"			release
 	"0.8.0.dev1"	development, see https://www.python.org/dev/peps/pep-0440
+    "0.8.0+dcl3"    local branch patches, before upstreaming
 
 	 NB: OPENQL_VERSION_STRING is also decoded by setup.py
 */
-#define OPENQL_VERSION_STRING "0.8.1.dcl3"
+#define OPENQL_VERSION_STRING "0.8.1+dcl3"
```
razvnane, line 8: This is still an illegal name. A legal name can be only 0.8.1.dev\<NUMBER\>. Adding other characters such as "+" or "dc" is not permitted. The only string allowed is "dev", so that is not an issue, i.e., no need to change the "." to "+"

wvlothuizen: thanks for your comment, I was looking for a way to give a meaningful version name to an upstream branch. This "+" syntax seemed to be allowed by the PEP, but I'll change to "dev" since it doesn't work

tests/test_skip.py
```diff
@@ -51,13 +51,12 @@ def test_skip_yes(self):
...
         qasm_rdr = ql.cQasmReader(starmon, prog)
-        qasm_str = f"""
-            version 1.0
+        qasm_str = f"""version 1.0
             qubits 7
             .{kernel_name}
             cnot q[2],q[5]
             skip 2
-            { x q[0] | y q[1] }
+            {{ x q[0] | y q[1] }}
```
razvnane, line 59: Is the double {} actually needed? My tests were working without these so is there a special reason for these?

.travis.yml
```diff
@@ -28,21 +28,23 @@ before_install:
...
+  - python3 -m pip install .
   - cd ../../.
-  - python3 -m pip install . 
+  - python3 -m pip install -v .
+
 script: python3 -m pytest
 
 
+# FIXME: moved to install, in search of problem that CI fails
+#script: python3 -m pytest
```
razvnane, line 47: clean this

tests/test_skip.py
wvlothuizen: I changed the string to a 'multi line f-string', which doesn't need an explicit "\n" to end lines, and uses {} to insert variable values (like in '{kernel_name}'). To escape "{" one must use "{{"

.travis.yml
wvlothuizen: removed

wvlothuizen:
> The cqasm_reader.cc update to put wait around each cnot/swap/cz, must go out or be put under an option.

is was already removed some time ago:

```
commit 5aaea50f1bd91a7ca3fd17f836695402f39ddac3
Author: QFer <f.j.grooteman@tudelft.nl>
Date:   Thu Apr 16 19:36:35 2020 +0200

    remove wait hack

```

PR #324, opened 2020-06-08
Visualizer

tmeer:
First commit of visualizer code. Not integrated yet, but can be ran seperately. The visualizer uses CImg.h for drawing to the screen, visualizer.h and visualizer.cpp contain the visualizer logic.


src/visualizer.cpp
```diff
@@ -0,0 +1,522 @@
...
+#include <CImg.h>
+#include <visualizer.h>
+#include <gate.h>
+
+using namespace cimg_library;
+using namespace ql;
+
+int main()
+{
+	// TODO: implement a generic grid structure object to contain the visual structure of the circuit, to ease positioning of components in all the drawing functions
```
razvnane: The main function can be removed

src/visualizer.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+	// Calculate amount of cycles.
+	unsigned int amountOfCycles = 0;
+	for (const gate* gate : gates)
+	{
+		const unsigned int gateCycle = (unsigned int)gate->cycle;
+		if (gateCycle > amountOfCycles)
+			amountOfCycles = gateCycle;
+	}
+	amountOfCycles++; // because the cycles start at zero, we add one to get the true amount of cycles
```
razvnane: this should be in a private helper function

src/visualizer.cpp
```diff
@@ -0,0 +1,522 @@
...
+	const gate* lastGate = gates.at(gates.size() - 1);
+	const unsigned int lastGateDuration = (unsigned int)lastGate->duration;
+	const unsigned int lastGateDurationInCycles = lastGateDuration / 40;
```
razvnane: what does 40 mean? Shouldn't this be a parameter read from somewhere?

src/visualizer.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+		//std::cout << "amount of cycles before compression: " << amountOfCycles << std::endl;
+		unsigned int amountOfCompressions = 0;
+		for (unsigned int i = 0; i < filledCycles.size(); i++)
+		{
+			//std::cout << i;
+			if (filledCycles.at(i) == false)
+			{
+				//std::cout << " not filled" << std::endl;
+				//std::cout << "\tcompressing... min cycle to compress: " << i - amountOfCompressions << std::endl;
```
razvnane: move all these commented prints to a DOUT call that can be enabled with the debug option selected

src/visualizer.h
```diff
@@ -0,0 +1,498 @@
...
+	ql::gate* gate8 = new ql::cphase(0, 2);
+	gate8->cycle = 9;
+	gates.push_back(gate8);
+
+	return gates;
+}
+
+std::vector<ql::gate*> getDemoCircuit1()
+{
+	std::vector<ql::gate*> gates;
```
razvnane, line 277: move these test/example cases to test_visualizer.py unit test and code the example in OpenQL and call the visualizer as pass

.gitignore
```diff
@@ -28,3 +28,7 @@ test_output
 docs/_build
 docs/_static
 docs/_templates
+
+# tim
+CMakeLists.txt
+.gitignore
```
razvnane, line 34: this should not be commited

CMakeLists.txt
```diff
@@ -61,6 +61,10 @@ SET(LIBQASM_SOURCE_DIRS
   "${PROJECT_SOURCE_DIR}/deps/libqasm/src/library"
 )
 
+SET(PYTHON_EXECUTABLE "/usr/bin/python3")
+SET(PYTHON_LIBRARY "/usr/lib/python3.6")
+SET(PYTHON_INCLUDE_DIR "/usr/include/python3.6")
+
```
razvnane, line 67: this should not be committed

src/passes.cc
```diff
@@ -293,6 +294,12 @@ void ReportStatisticsPass::runOnProgram(ql::quantum_program *program)
     ql::report_statistics(program, program->platform, "todo-inout", getPassName(), "# ");
 }
 
+void VisualizerPass::runOnProgram(ql::quantum_program *program)
+{
+    Layout layout;
```
razvnane, line 299: where is this initialized?

src/visualizer.cpp
tmeer: Removed

src/visualizer.cpp
tmeer: Done

src/visualizer.cpp
tmeer: Yes, this is supposed to be read from somewhere. For now it is a magic number that defines the duration of one cycle.

src/passes.cc
tmeer: It is initialized with default values when you create a Layout object. In the future the default values will instead be read from a configuration file, and can also be changed by including a Layout object in the pass options.

setup.py
```diff
@@ -29,7 +29,7 @@
     cmd = 'cmake ..'
     proc = subprocess.Popen(cmd, shell=True)
     proc.communicate()
-    cmd = 'make -j{}'.format(nprocs)
+    cmd = 'make VERBOSE=1 -j{}'.format(nprocs)
```
razvnane, line 32: no verbose

CMakeLists.txt
```diff
@@ -85,12 +89,12 @@ INCLUDE_DIRECTORIES(
...
   ${LIBQASM_INCLUDE_DIRS}
-#  ${EIGEN_INCLUDE_DIRS}
-#  ${EIGEN_UNSUPPORTED_INCLUDE_DIRS}
+  ${EIGEN_INCLUDE_DIRS}
+  ${EIGEN_UNSUPPORTED_INCLUDE_DIRS}
   ${PROJECT_SOURCE_DIR}/src
 )
-
 IF(CMAKE_COMPILER_IS_GNUCXX)
+  LINK_LIBRARIES("-lX11")
```
razvnane, line 97: this should not be here, but in SWIG_LINK_LIBRARIES in swig folder. Please check how to define a X11 library following the example of lemon. Also make cmake search for the library and give a configuration error if not found.

src/passes.cc
```diff
@@ -293,6 +294,14 @@ void ReportStatisticsPass::runOnProgram(ql::quantum_program *program)
     ql::report_statistics(program, program->platform, "todo-inout", getPassName(), "# ");
 }
 
+void VisualizerPass::runOnProgram(ql::quantum_program *program)
+{
```
razvnane, line 298: Please add doxygen comments similar to the other passes.

src/visualizer.cc
```diff
@@ -0,0 +1,539 @@
...
+			(layout.grid.cellSize - node.radius * 2) / 2,
+			labelColumnWidth,
+			cycleNumbersRowHeight,
+			(unsigned int)gate->cycle,
+			operandType == CLASSICAL ? (unsigned int)gate->creg_operands.at(index) + circuitData.amountOfQubits : (unsigned int)gate->operands.at(index)
+		};
+
+		switch (node.type)
+		{
+			case NONE:		break; // Do nothing.
```
razvnane, line 430: Add warning message that NONE type encountered

src/visualizer.cc
```diff
@@ -0,0 +1,539 @@
...
+			case GATE:		drawGateNode(image, layout, circuitData, node, positionData); break;
+			case CONTROL:	drawControlNode(image, layout, circuitData, node, positionData); break;
+			case NOT:		drawNotNode(image, layout, circuitData, node, positionData); break;
+			case CROSS:		drawCrossNode(image, layout, circuitData, node, positionData); break;
+		}
```
razvnane, line 435: add default branch to switch statement with warning message for the user.

PR #327, opened 2020-06-10
Fix Windows compilation with MSVC

razvnane:
This PR fixes the broken compilation on Windows due to the addition of the eigen library needed by the unitary decomposition as well as to some inconsistencies in how platform strings are accessed (done recently to clean up the backend code).


imranashraf:
I will also suggest to use [Chocolately ](https://chocolatey.org/) for setting up software on windows. This will make it very easy to setup development software on windows as everything can be scripted. for instance, gcc (MinGW) can be installed with a single command:

`choco install mingw`

This will also result in a lot less space requirement on windows.

PR #329, opened 2020-06-11
Eigen compilation time

anneriet:
This reduces the compilation time by reducing the number of Eigen header files in the OpenQL repository, through using a sparse-checkout for the Eigen submodule. To make it somewhat usable, the necessary commands for the sparse-checkout are in a shell script, and how to use it is put in the documentation under the "installation" instructions.


razvnane:
Please change this PR to develop and make sure first all tests pass. We can import your fixes into modular after that.

PR #331, opened 2020-06-15
[WIP] test CI

wvlothuizen:
this is to test whether CI also fails on (a copy of) branch develop


razvnane:
What does this PR modifies? I see it pointing inside deps/ but what is the reason for that? Shouldn't this PR made to libqasm instead?

wvlothuizen:
I had trouble with CI on PR #321  which I thought might stem from development itself, so this was just to trigger CI on (basically) development.
It builds fine, so I'll have to look elsewhere, and will close this PR.

PR #338, opened 2020-07-08
Add doxygen to documentation build

jvanstraten:
Since I figured out how to do C++ documentation generation on ReadTheDocs for libqasm anyway, I figured I'd copypaste the logic into OpenQL as well. The generated docs are only as clear and organized as the source code is, but it's still nice to be able to click through the files and class hierarchy outside of an IDE.

@razvnane since I don't have access to the RTD interface, can you trigger a build on this branch sometime to make sure it works? All of this worked in one go on libqasm so I don't see why it shouldn't, but you never know.


docs/conf.py
```diff
@@ -14,6 +14,22 @@
...
 
+# -- Doxygen build -----------------------------------------------------------
+import subprocess
+import os
+import sys
+original_workdir = os.getcwd()
+try:
+    docs_dir = os.path.dirname(__file__)
+    os.chdir(docs_dir)
+    if not os.path.exists('doxygen/doxy'):
```
razvnane, line 25: why do we need this check?

jvanstraten: I suppose you already merged, but I'll comment anyway; it's there because conf.py is called twice by RTD, and Doxygen has no incremental build support. So it saves some build time.

PR #339, opened 2020-07-09
Bump libqasm to new version

jvanstraten:
This updates the libqasm subrepo to the latest version and updates the build system and test cases accordingly to make things work. The cQASM reader still uses the original libqasm API through its compatibility layer.


tests/test_skip.py
```diff
@@ -55,7 +55,7 @@ def test_skip_yes(self):
                    "qubits 7\n"                 \
                    "." + kernel_name + "\n"     \
                    "cnot q[2],q[5]\n"           \
-                   "skip 2"                     \
+                   "skip 2\n"                   \
                    "{ x q[0] | y q[1] }\n"
```
razvnane, line 59: Please pull develop into your branch first.  This test case changed in a recent PR. It is strange though why this merge request does not point to that merge but to an older version of the target file.

jvanstraten: I probably did `git pull` `git merge develop` instead of `git fetch` `git merge origin/develop` so I merged some in-between version I guess. Bad habit.

swig/CMakeLists.txt
```diff
@@ -1,17 +1,16 @@
 FIND_PACKAGE(SWIG REQUIRED)
 INCLUDE(${SWIG_USE_FILE})
 
-#FIXME: duplicates parent
```
jvanstraten, line 4: To clarify, someone removed this python dependency from the swig subdir CMakeLists in response to the same piece of code being unnecessarily copypasted from libqasm into the main CMakeLists. Now that those libqasm's CMakeLists are more or less sane and just require an `add_subdirectory()` I'd removed the unnecessarily copypasted bit from the main CMakeLists, so stuff broke. I opted to just go back to the original state before libqasm was pulled in.

razvnane: Wouter removed this after a painfully long process to debug a failed CI due to this code selecting a different python interpreter than the parent one. So let's see if with the modifications in the other files this now passes CI.

jvanstraten: That's... odd. Looks like it passes again now though.

razvnane: Yes, that's indeed odd. You can check the PR #321 if you are interested in that problem. Good that it passes again, I will merge.

PR #342, opened 2020-07-20
Update CMakeLists.txt

mmoelle1:
Using CMAKE_SOURCE_DIR does not work correctly if OpenQL is used as subproject within another project. Either use CMAKE_CURRENT_SOURCE_DIR or (better choice) PROJECT_SOURCE_DIR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,11 +46,11 @@ SET(CLI11_INCLUDE_DIRS
 )
 
 SET(EIGEN_INCLUDE_DIRS
-  "${CMAKE_SOURCE_DIR}/deps/eigen/Eigen"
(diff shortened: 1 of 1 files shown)


jvanstraten:
This and many more issues with the build system will already be taken care of by #340. CMake-wise it should be done already, I'm just working on conda still since that supposedly worked at some point according to the docs. It'd be great if you could try whatever you're doing with that branch; the more it's tested the better.

razvnane:
As Jeroen pointed out earlier, this issue should already be fixed by PR #340 (now merged), so I will close this. Nevertheless, in case this issue somehow will still cause problems, please reopen or create a new issue.

Issue #347, opened 2020-09-02
Tutorials broken due to differently-named qasm output file


jvanstraten:
The tutorials expect a `<program-name>.qasm` file to be written to the output directory. Apparently that's not true anymore; now it writes a few differently named qasm files for the various passes. This is IMO a breaking change. Why doesn't it just output the final qasm file after all passes with the original naming?

@razvnane I'm assuming this is due to the modularization changes? Is this a permanent thing (i.e. docs should be updated) or should the code be fixed?

PR #348, opened 2020-09-02
Documentation improvements

jvanstraten:
Documents how to enable initial placement (and use the other options) during installation pending a resolution for GLPK, and improves the tutorials somewhat.

~~Based on the outcome for #347 this may receive another commit to fix the docs, hence draft status.~~

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -19,6 +19,12 @@ Installing the pre-built package
 
 Pre-built packages are available for OpenQL.
 
+.. note::
+
+    Initial placement is currently not included with the binary distribution
+    due to a license incompatibility with one of the dependencies (GLPK, to
+    be specific). Until this is resolved somehow, you'll have to build from
+    source if you need it.
 
 
 Pre-built Wheels
@@ -88,6 +94,7 @@ The following packages are required to compile OpenQL from sources:
 - [Optional] pytest used for running tests
 - [Optional] Graphviz Dot utility to convert graphs from dot to pdf, png etc
 - [Optional] XDot to visualize generated graphs in dot format
+- [Optional] GLPK if you want initial placement support
 
 
 Notes for Windows Users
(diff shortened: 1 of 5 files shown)


docs/installation.rst
```diff
@@ -157,7 +171,7 @@ OpenQL sources for each release can be downloaded from github `releases <https:/
 
 ::
 
-    git clone https://github.com/QE-Lab/OpenQL.git
+    git clone https://github.com/QE-Lab/OpenQL.git --recursive
```
razvnane, line 174: why the new flag?

docs/installation.rst
```diff
@@ -169,16 +183,34 @@ Running the following command in the python (virtual) environment in Terminal/Po
 
     cd OpenQL
     git submodule update --init --recursive
-    python setup.py install
+    pip install -v
```
razvnane, line 186: why change the python setup.py to pip install ?

docs/installation.rst
```diff
@@ -169,16 +183,34 @@ Running the following command in the python (virtual) environment in Terminal/Po
...
 
 Or in editable mode by the command:
 
 ::
 
-    pip install  -e .[develop]
+    pip install -v -e .
+
+Editable mode has the advantage that you'll get incremental compilation if you ever change OpenQL's C++ files, but it's
+a bit more fragile in that things will break if you move the OpenQL repository around later.
```
razvnane, line 195: so when you move the repo, clean first, that's what you mean?

docs/qx_example.rst
```diff
@@ -1,7 +1,7 @@
 QX Simulation
 =============
 
-This tutorial explains how to compile an OpenQL program and execute it on QX. We will use the example of rolling an 8-faced dice. Rolling this dice results in 1 out of 8 outcomes. The complete code for this example is available in examples/dice.py
+This tutorial explains how to compile an OpenQL program and execute it on QX. We will use the example of rolling an 8-faced dice. Rolling this dice results in 1 out of 8 outcomes. The complete code for this example is available in ``examples/dice.py``, but you can also copy the snippits over to your own script as we walk through it.
```
razvnane, line 4: Last sentence after 'but' can be left out. IMO it does not add any new information.

docs/installation.rst
jvanstraten: If you just do `--recursive` during pull you don't have to do `git submodule update --init --recursive`. It's redundant, but so many of the errors I've heard from people so far are just due to them not doing the latter I figured some redundancy wouldn't hurt.

docs/installation.rst
jvanstraten: Because the former has been part of the issues I've been having with Nikiforos, I think. If you install using `setup.py` directly pip's package management logic is bypassed so pip may or may not be aware of the installation. That's literally the only difference with `pip install`, since `pip` just calls `setup.py`.

docs/installation.rst
jvanstraten: Also, about the "having to clean after move" thing CMake throws at you, I already built a thing into `setup.py` to check for this and clean automatically to avoid that silly and confusing error.

razvnane (approved):

PR #352, opened 2020-09-04
Reduce frequency of CI failures due to sourceforge

jvanstraten:
The chocolatey install of winflexbison fails way too often, due to sporadic download errors from SourceForge. This tries to reduce the frequency of these false negatives (in an admittedly ugly way, because chocolatey doesn't support retries on its own).


jvanstraten:
It looks like sourceforge fixed things on their end, because it hasn't happened in a while, and I can't really be bothered learning powershell for this unless it becomes an issue again.

PR #353, opened 2020-09-05
Fix cc files

jvansomeren:
.cc programs in tests and in examples that were not working because of use of deprecated features, were updated.
In the related CMakeLists.txt files, these were uncommented in.
make test passes now for all


jvanstraten:
Can you remove the fixme comments in the tests/examples CMakeLists as well? Since they don't apply anymore. Looks good to me otherwise.

jvansomeren:
Removed FIXME lines.
Current failures don't seem to be caused by fixes but by make system.

jvanstraten:
>Current failures don't seem to be caused by fixes but by make system.

No, it's caused by you rewinding libqasm back to before the rewrite. Please don't add+commit changes unrelated to the pull request for one thing, separate features/bugfixes/updates/whatever belong in separate PRs. As for why this happened in the first place, when you do `git checkout <branch>` to switch branches, git for some reason doesn't update submodules as well. You have to do `git submodule update` for that. Because you never did that when the actual new version of libqasm was committed into OpenQL, libqasm was still stuck on an old version from way back on your system. By add+committing libqasm rather than doing `git submodule update` you ended up rewinding libqasm back to that old version.

For reference, to actually pull new changes from a submodule, you need to do `git pull` from within the submodule, then `git checkout` for whatever branch or commit you want to set it to, and then you add+commit in the parent repository. You can never get new changes without pulling in the submodule first. But such things (along with any fixes needed) belong in a seperate pull request.

jvanstraten (approved):

razvnane (approved):

PR #354, opened 2020-09-09
Visualizer

tmeer:
Resolved Jeroen's comments from last pull request:

- Reverted back to the default CI compiler and instead updated code to no longer use bugged language construct.
- Removed the visualizer test (for now).
- X11 is now linked in the correct place, and if X11 is not found on Linux or Mac the visualizer disables itself. On Windows no additional library is necessary to run the visualizer, therefore the visualizer is always enabled.
- All the checks now pass.
- Seperated the visualizer's internals (including CImg.h) from its outward facing interface.
- Checked CImg's license, the header file is dual-licensed and can be licensed under CeCILL-C, which is compatible with OpenQL's license. No other files from CImg are used.

**Visualizer usage**

The visualizer can be ran by adding the visualizer pass to the compiler and compiling the quantum program you want to run. Python example:
```
c = ql.Compiler("example_compiler")
c.add_pass("Visualizer");
c.set_pass_option("Visualizer", "visualizer_config_path", "full_path_to_file"));

# add program and gates

c.compile(program)
```
The general visualization parameters are read from a configuration file. The path to this file should be added through a pass option for the visualizer as seen above. If a parameter is missing from that file, the default hardcoded value contained in `src\visualizer.h` will be used instead.

**Custom gates**

When using custom gates the default gate visualizations are not used and the visualization needs to be defined by the user. Take for example the following custom Hadamard gate defined on qubit 1:
```
"h q1": {
  "duration": 40,
  "latency": 0,
  "qubits": ["q1"],
  "matrix": [ [0.0,1.0], [1.0,0.0], [1.0,0.0], [0.0,0.0] ],
  "disable_optimization": false,
  "type": "mw",
  "cc_light_instr_type": "single_qubit_gate",
  "cc_light_instr": "h",
  "cc_light_codeword": 91,
  "cc_light_opcode": 9,
  "visual": {
    "connectionColor": [0, 0, 0],
    "nodes": [
      {
        "type": "GATE",
        "radius": 13,
        "displayName": "H",
        "fontHeight": 13,
        "fontColor": [255, 255, 255],
        "backgroundColor": [70, 210, 230],
        "outlineColor": [70, 210, 230]
      }
    ]
  }
}
```
Note the new "visual" attribute, which defines the look of the custom gate. Each gate has a `connectionColor` which defines the color of the connection line for multi-operand gates, and an array of 'nodes'. A node is the visualization of the gate acting on a specific qubit or classical bit. If a Hadamard gate is acting on qubit 3, that is represented by one node. If a CNOT gate is acting on qubits 1 and 2, it will have two nodes, one describing the visualization of the CNOT gate at qubit 1 and one describing the visualization on qubit 2. A measurement gate measuring qubit 5 and storing the result in classical bit 0 will again have two nodes.

Each node has several attributes describing its visualization:
- `type` : the visualization type of the node, see below for a list of the available types
- `radius` : the radius of the node in pixels
- `displayName` : text that will be displayed on the node (for example 'H' will be displayed on the Hadamard gate in the example above)
- `fontHeight` : the height of the font in pixels used by the `displayName`
- `fontColor` : the color of the font used by the `displayName`
- `backgroundColor` : the background color of the node
- `outlineColor` : the color of the edge-line of the node

The colors are defined as RGB arrays: `[R, G, B]`.

The type of the nodes can be one of the following:
- `NONE` : the node will not be visible
- `GATE` : a square representing a gate
- `CONTROL` : a small filled circle
- `NOT` : a circle outline with cross inside (a CNOT cross)
- `CROSS` : a diagonal cross

When a gate has multiple operands, each operand should have a node associated with it. Simply create as many nodes in the `nodes` array as there are operands and define a type and visual parameters for it. Don't forget the comma to seperate each node in the array.

**Future work**

Features and issues on the todo-list are:
- display wait/barrier gates (not possible right now because the program passed to the visualizer does not contain these gates anymore)
- gate connections overlap when in the same cycle
- add the classical bit number to the measurement connection when classical bit lines are grouped
- add a proper measurement symbol
- add an option to save the image and/or only generate that image without opening a window
- add option to represent each gate as a pulse instead of an abstract symbol


.github/workflows/test.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - develop
+    - visualizer
```
jvanstraten, line 7: This doesn't really belong here. To get CI to run while you're still developing, you can just make a draft PR. Note that you can continue to commit to both draft and non-draft PR branches after they're opened to resolve issues, you don't need to close and then reopen later.

jvanstraten (changes requested):
Nice, looks good to me now aside from the explicit addition of the visualizer branch to the github actions CI logic. I'll accept when that line is removed. (e-mail me when you do or request review from me, otherwise I don't think I'll get notified and I might forget)

.github/workflows/test.yml
jvansomeren: I have not checked the code; trust that that is ok.
Have only read the readme.
I'm glad that we reached this milestone and propose to continue with the merge.

jvansomeren:
I have not checked the code; trust that that is ok.
Have only read the readme.
I'm glad that we reached this milestone and propose to continue with the merge.

.github/workflows/test.yml
tmeer: Removed this line. Also removed some other personal development lines from .gitignore.

jvanstraten (approved):

PR #359, opened 2020-10-12
Take cbox out

jvansomeren:


jvansomeren:
The removed tests didn't output qasm or qisa to be checked against golden versions.
So the current pull request is complete; it doesn't make files in golden unused.

jvanstraten (approved):

PR #360, opened 2020-10-12
upstream CC backend changes (new)

wvlothuizen:


jvansomeren (changes requested):
When the scheduler reports after Dependence Graph construction, that it is not a DAG, there is a problem that should be debugged. There is a reasonable chance that also for other programs the dependence graph is not correct and that goes unnoticed. So I would advise against accepting this pull request.

jvanstraten:
The libqasm submodule also got downgraded, causing literally all builds to fail. Please run

```
cd deps/libqasm
git fetch
git checkout f10b04fb
cd ../..
git add deps/libqasm
```

and commit.

In the future, when git says a submodule has "new commits" after a merge (or pull), run `git submodule update --recursive`. git is extremely stupid about this; the "new commits" it's referring to actually means "old commit" in this case, because the merge/pull doesn't update your working tree automatically for submodules like it does for files. So if you `git add` it to your commit you force it back to the previous state from your working tree. `git submodule update --recursive` aligns your working tree with the merged state. Why git doesn't do this automatically I'll never know, but googling I did find [this](https://gist.github.com/ejmr/453edc19dd596e472e90) to automate what IMO should be default behavior.

jvanstraten:
Also, where did `src/CMakeLists.txt` come from? You should be building from the root directory (which doesn't work right now as evidenced by CI because libqasm was reverted to an ancient version).

I know by the way that normally all subdirectories have CMakeLists in them, it just didn't make sense to me for this particular project structure. In any case, if you think that should be revised, it should be through a different PR.

wvlothuizen:
@Hans: I did observe the DAG error once, and just quickly inserted a FIXME to allow reproducing. And I replaced the warning by Fatal, because it does not seem to make sense to continue. It would be better to make a Github issue on this, and I will after my laptop has recovered from sudden death (which may very well take a few days). Still, I don't think the actual problem is caused by this PR, somehow one of the PycQED test cases triggered this behaviour (and may still do so).

wvlothuizen:
@Jeroen: I'll clean up the mess once my laptop works again, I did see a libqasm conflict and performed a "git submodule update", but without the "--recursive". 
By the way, great that we now have CI on PRs.
I'll request a review once I remove the "WIP" status of this PR.

wvlothuizen:
Tests temporarily disabled:

- [ ] test_cc.cc::test_qi_example

- [ ] codegen_cc.cc : " verify dimensions"

wvlothuizen:
@hans: I filed issue #363 regarding the scheduler that seems to block this PR

wvlothuizen:
ready for merge, note that I'll continue to properly call scheduler

jvansomeren (approved):

PR #361, opened 2020-10-12
Multi core

jvansomeren:
I suspect most of the tests in develop test the wrong qasm.
The tests that also execute the backend, must test a qasm that is generated just before/after code generation.
A pass generating this was not present until now, so I added it.
It generates ’progname’_last.qasm.

See the new tests/test_multi_core.py in the multi_core branch for an example of its use.
I may be wrong with names etc. but the idea and the problem are clear, I hope.

So this means that we have to regenerate golden qasm outputs
when the qasm just before/after code generation is meant.
The current _scheduled.qasm is just before entering code generation or entering QX.

See the multi-core branch for the necessary changes.
I didn’t update the other tests.

The multi-core branch is ok, all tests pass there including the multi-core test itself.


jvanstraten (approved):
I'm not sure why the mapper is *this* different from the previous version, it seems like all kinds of things unrelated to multi-core were changed. But whatever, as far as I'm concerned the mapper code is entirely your domain.

PR #370, opened 2020-11-02
Feedback support for CC backend

wvlothuizen:


wvlothuizen:
@jvansomeren @jvanstraten :
Ready for review in my opinion, merged with develop and all checks pass

jvanstraten (commented):
In terms of functionality, I'm of the opinion that if CI passes it's good enough to merge into develop by definition, so I've mostly just scrolled through the changes to look for anything that might stand out. The only thing I noticed is that indentation and other whitespace-related style is very inconsistent? I think I've seen anywhere between 4 and 7 spaces being used. More consistent usage of the utils library would also be nice, the CC backend being the only piece of code left that doesn't use it exclusively.

I can change those things for you if you want, considering I'm the one most picky about it. I only refrained from touching CC before (except where necessary to solve compile errors) to prevent huge merge conflicts. It shouldn't take me too long to do and I don't mind doing it.

wvlothuizen:
I'll have a look at my whitespace settings: I recently switched to CLion which may have upset things.

I'm fine with you making the edits you propose, shall we use a separate PR for that so I can have a glance at the changes?

jvanstraten:
Sure, I'll do a pass over it on Monday and PR it into DiCarloLab-Delft:feature/cc/feedback.

wvlothuizen:
@jvansomeren @jvanstraten :

do you think can we merge, the refactoring by Jeroen is in place, and the tests only seem to fail because the Windows build fails to obtain Swig/Bion

jvanstraten:
Oops, I forgot about that. I restarted CI to see if it was a temporary remote issue and fixes itself, but if not I'll just override it/merge anyway and make a note of having to fix it on develop.

jvanstraten (approved):

PR #372, opened 2020-11-09
fix for Python include path on windows (by JvS)

wvlothuizen:
please delay merging until we tested on a real setup at DCL


jvanstraten (approved):

jvanstraten:
WIP tag but not a draft PR? Are you still intending to change something?

wvlothuizen:
I thought to delay until we tested on a real setup (which uses venv), but we could also start a new PR if necessary

PR #373, opened 2020-11-09
Extending the dependence graph with improved support for gate commutation

jvansomeren:
Initially this branch started off to create a new pass relying on the dependence graph and on the resource-constrained scheduler to exploit commutation rules of CNOTs and of CZs to reorder those inside each set of commutable gates; the pass generates all possible variations and finds the best schedule reducing circuit latency. This pass has been added.
However while doing, the commutation rules themselves got under investigation: they could be improved and even be extended with rules concerning arbitrary angle X and Z rotations; and their representation in the dependence graph could be improved: more readable, making the state machines explicit, and removing some unnecessary encodings in the implementation, i.e. a cleanup of the code. The end-result is a largely improved dependence graph implementation, both in readability and in power. For testing, many tests needed to be added to gain confidence in the new implementation.


jvanstraten (approved):

jvanstraten (approved):
I haven't looked at the dependency analysis code enough to review in-depth, but the code looks good enough to me.

PR #379, opened 2020-11-26
[WIP] Conditional execution

jvanstraten:
Tracks the condex branch, where conditional execution support is being worked on.


jvansomeren:
Condex is functionally complete, all passes were updated to support conditional execution.
Now waiting whether all CI tests pass.

src/kernel.cc
```diff
@@ -741,33 +783,101 @@ void quantum_kernel::gate(const Str &gname, UInt q0, UInt q1) {
...
+    auto lcregs = cregs;
+    auto lbregs = bregs;
+    gate_add_implicits(gname, lqubits, lcregs, duration, angle, lbregs, gcond, gcondregs);
+    if (!gate_nonfatal(gname, lqubits, lcregs, duration, angle, lbregs, gcond, gcondregs)) {
+        QL_FATAL("Unknown gate '" << gname << "' with qubits " << lqubits);
+    }
+}
+
+/**
+ * preset condition to make all future created gates conditional gates with this condition
```
wvlothuizen, line 838: This looks like a hack to me, wouldn't it be better to add the support for conditional gates where needed

src/kernel.cc
```diff
@@ -741,33 +783,101 @@ void quantum_kernel::gate(const Str &gname, UInt q0, UInt q1) {
...
+ * clear preset condition again
+ */
+void quantum_kernel::gate_clear_condition() {
+    gate_preset_condition(cond_always, {});
+}
 
-    if (!gate_nonfatal(gname, qubits, cregs, duration, angle)) {
-        QL_FATAL("Unknown gate '" << gname << "' with qubits " << qubits);
+/**
+ * add implicit parameters to gate to match IR requirements
```
wvlothuizen, line 863: I don't understand the comment. Most parameters are unused (except for logging). Should probably be private in .h file

src/kernel.h
```diff
@@ -30,20 +31,24 @@ class quantum_kernel {
...
     utils::UInt             qubit_count;
     utils::UInt             creg_count;
+    utils::UInt             breg_count;
     kernel_type_t           type;
     circuit                 c;
     utils::Bool             cycles_valid; // used in bundler to check if kernel has been scheduled
     utils::Opt<operation>   br_condition;
     utils::UInt             cycle_time;   // FIXME HvS just a copy of platform.cycle_time
     instruction_map_t       instruction_map;
+    utils::Vec<utils::UInt> cond_operands;    // see gate interface: condition mode to make new gates conditional
```
wvlothuizen, line 41: parameters of quantum_kernel::gate() were added at the end to maintain backward compatibility, but internally I would maintain a more natural order.

src/kernel.h
```diff
@@ -154,7 +172,25 @@ class quantum_kernel {
...
+        const utils::Vec<utils::UInt> &bregs = {},
+        cond_type_t gcond = cond_always,
+        const utils::Vec<utils::UInt> &gcondregs = {}
+    );
+    void gate_preset_condition(
+        cond_type_t gcond,
+        const utils::Vec<utils::UInt> &gcondregs
+    );
+    void gate_clear_condition();
+    void gate_add_implicits(
```
wvlothuizen, line 185: Should probably be private

src/openql_i.cc
```diff
@@ -257,19 +265,144 @@ void Kernel::gate(
...
     kernel->gate(name, {qubits.begin(), qubits.end()}, {(destination.creg)->id} );
 }
 
+void Kernel::gate_preset_condition(
+    const std::string &condstring,
+    const std::vector<size_t> &condregs
+) {
+    QL_DOUT("Python k.gate_preset_condition("<<condstring<<", condregs)");
+    ql::cond_type_t condvalue;
+    if      (condstring == "COND_ALWAYS") condvalue = ql::cond_always;
```
wvlothuizen, line 339: move repeated code block to function

src/openql_i.cc
```diff
@@ -257,19 +265,144 @@ void Kernel::gate(
...
+    else if (condstring == "COND_AND") condvalue = ql::cond_and;
+    else if (condstring == "COND_NAND") condvalue = ql::cond_nand;
+    else if (condstring == "COND_OR") condvalue = ql::cond_or;
+    else if (condstring == "COND_NOR") condvalue = ql::cond_nor;
+    else if (condstring == "COND_XOR") condvalue = ql::cond_xor;
+    else if (condstring == "COND_NXOR") condvalue = ql::cond_nxor;
+    else {
+        throw std::runtime_error("Error: Unknown condition " + condstring);
+    }
+    kernel->gate(
```
wvlothuizen, line 394: add kernel->condgate(), so C++ users have similar API

src/program.cc
```diff
@@ -109,31 +111,31 @@ void quantum_program::add_program(const quantum_program &p) {
 
 void quantum_program::add_if(const quantum_kernel &k, const operation &cond) {
     // phi node
-    quantum_kernel kphi1(k.name+"_if", platform, qubit_count, creg_count);
+    quantum_kernel kphi1(k.name+"_if", platform, qubit_count, creg_count, breg_count);
```
wvlothuizen, line 114: adding breg_count makes sense, but I actually never understood why a kernel has its own qubit_count and creg_count, separate from a program

wvlothuizen (changes requested):
Looks good overall, although I have some comments. Documentation update will be a separate PR I assume?

python/openql.i
```diff
@@ -545,9 +547,15 @@ arg1 : str
 arg2 : []
     list of qubits
 arg3 : int
-    duration in ns (at the moment it is only supported for wait instruction, in the future it will be extended to override duration of any gate)
-arg3 : double
-    angle of rotation, used internally only for rotations (rx, ry and rz)
+    duration in ns (default: 0)
+arg4 : double
+    angle of rotation, used internally only for rotations (rx, ry and rz) (default: 0.0)
+arg5 : []
```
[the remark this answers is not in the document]
jvansomeren: I'll add condgate in C++ interface as in Python interface; I forgot to add it.
But will keep the current gate() interface as well because an all-inclusive interface is the ultimate fallback
and having all kinds of variations I want to avoid.

src/gate.h
```diff
@@ -167,19 +167,29 @@ const utils::Complex nop_c[] = {
...
 
-
+/*
+ * additional definitions for describing conditional gates
+ */
+typedef enum e_cond_type {
+    // 0 operands:
+    cond_always, cond_never,
+    // 1 operand:
+    cond, cond_not,
```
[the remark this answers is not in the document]
jvansomeren: True, stumbled several times about it when updating the code.
Do you have a name proposal?
cond_unary, cond_cond, cond_single, cond_one, cond_var, ...?

src/kernel.cc
```diff
@@ -105,16 +109,22 @@ void quantum_kernel::h(UInt qubit) {
 
 void quantum_kernel::rx(UInt qubit, Real angle) {
     c.push_back(new ql::rx(qubit,angle));
+    c.back()->condition = condition;
```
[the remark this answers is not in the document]
jvansomeren: I didn't want to touch those: Those interfaces I consider deprecated (i.e. new ql::rx(...), and similar).
And I don't really understand them.
These will be dropped when further cleaning up the IR.
But with the omnipotent gate() interface, we could call here gate() as well.

src/kernel.cc
```diff
@@ -462,6 +482,15 @@ Bool quantum_kernel::add_default_gate_if_available(
     }
 
     if (result) {
+        c.back()->breg_operands = bregs;
+        if (gcond != cond_always && is_non_conditional_gate ) {
+            QL_WOUT("Condition " << gcond << " on default gate '" << gname << "' specified while gate cannot be executed conditionally; condition will be ignored");
```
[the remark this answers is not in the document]
jvansomeren: Being forgiving and intelligent on this also is a 'good thing'.
For example, not that I like it, but with 'wait' or 'barrier' in a decomposition of cz, and then creating that cz with a condition
would stumble into this.

src/kernel.cc
```diff
@@ -741,33 +783,101 @@ void quantum_kernel::gate(const Str &gname, UInt q0, UInt q1) {
...
  * return the gate (or its decomposition) by appending it to kernel.c, the current kernel's circuit
  */
 void quantum_kernel::gate(
     const Str &gname,
     const Vec<UInt> &qubits,
     const Vec<UInt> &cregs,
     UInt duration,
-    Real angle
+    Real angle,
+    const Vec<UInt> &bregs,
```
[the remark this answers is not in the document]
jvansomeren: See above, will add condgate, will keep gate as omnipotent fall-back.

src/kernel.cc
jvansomeren: It is meant for gates that are decomposed by passes instead of steered by the config file's decompositions,
e.g. toffoli_decomposition and k.clifford(). I didn't want to clutter up those codes with conditional execution
but instead to do it under the hood.
We could argue about having it as a Python interface.

src/kernel.cc
jvansomeren: Was meant as a place holder to expand gates without arguments as well.
It is really ugly to have this now implicit in the IR; see e.g. src/scheduler.cc when creating depgraph in Init().
And don't want to update the prototype each time we add an other translation.

src/kernel.h
jvansomeren: Yes, true, will have a look.
Although with cleaning up the IR, this will change anyhow.

src/kernel.h
jvansomeren: gate_add_implicits must be private, the clear_condition not since it is an interface; see above.

src/openql_i.cc
jvansomeren: Yes, could have a try, although openql_i.cc is not a usual C++ implementation file, so other rules hold.
So must probably factor this code out and put it in src/kernel.*

src/openql_i.cc
jvansomeren: Yes.

src/program.cc
jvansomeren: Likewise, but suppose you have two independent kernels and want to synthesize them in parallel.
That was the only example I could come up with.
Not convincing though.
It is also very confusing already to have those separate from program and also separate from platform.

src/kernel.h
jvansomeren: Propose to leave them there.
These are not the arguments of a gate (see src/gate.h, there I rearranged the fields) but
meant to keep the preset_condition (see the related interface).

python/openql.i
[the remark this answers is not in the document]
jvansomeren: Added condgate.

src/gate.h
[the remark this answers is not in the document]
jvansomeren: Chose cond_unary

jvansomeren:
Hi Wouter, could you have a look again?

src/gate.h
```diff
@@ -167,19 +167,31 @@ const utils::Complex nop_c[] = {
...
 
-
+/*
+ * additional definitions for describing conditional gates
+ */
+typedef enum e_cond_type {
+    // 0 operands:
+    cond_always, cond_never,
+    // 1 operand:
+    cond_unary, cond_not,
```
wvlothuizen, line 177: giving it some more thought, it remains tough: "cond" is a prefix, and the second part is a Boolean operation. In that sense the original "cond" is correct, but on it's own it doesn't 'feel' logical. "cond_unary" uses a different system, and "not" is also unary. 
I feel that the Python interface could better be without the "COND"-part, and then an empty string as operation does make sense.

wvlothuizen (approved):

jvansomeren:
It is ok to merge

jvansomeren (approved):

PR #381, opened 2020-11-30
Update cQASM reader to new libqasm API

jvanstraten:
Updates the cQASM reader to use the new libqasm API, and enable cQASM 1.1 features (such as variables). Includes commented-out support for bit registers that will be added to the OpenQL IR in #379. Also, unlike the previous version, the new version allows the libqasm gateset and rules for converting to OpenQL IR to be configured using a secondary JSON file.


src/utils/opt.h
```diff
@@ -60,8 +60,8 @@ class Opt {
      * trick std::optional uses to avoid this is not implemented here.
      */
     template<class Arg1, class... Args>
-    Opt(Arg1 arg1, Args&&... args) {
-        emplace(arg1, std::forward<Args>(args)...);
+    Opt(Arg1 &&arg1, Args&&... args) {
+        emplace(std::forward<Arg1>(arg1), std::forward<Args>(args)...);
```
mbrobbel, line 64: 🐠

jvansomeren (approved):

PR #390, opened 2021-02-18
Mapping graph visualizer

tmeer:
The mapping graph visualizer provides a way to follow the movements of the virtual qubits throughout the execution of the circuit.

It does so by adding an additional struct parameter to the definition of a gate in gate.h. When the mapper adds a swap gate, this struct is filled with the real and virtual qubits involved in the swap. This struct is then propagated through all the gates the swap decomposes into, thereby making each gate involved in the swap identifiable. The visualizer then uses this information to construct the mapping graph.

Note that the swap information is _not_ propagated through a clifford optimization. Doing so would contaminate gates outside of a swap that are optimized together with gates from inside a swap with the swap parameters. This does not have any effect on the visualization of the mapping graph.


jvanstraten (approved):
Looks good to me.

jvanstraten:
Hm. Not sure why this wasn't merged, I guess I expected someone else to look at it too. An easy-fix merge conflict snuck in, so I fixed that. Will merge when CI finishes.

PR #395, opened 2021-03-22
[WIP] add rc scheduler for CC

wvlothuizen:


jvansomeren (commented):
I noted that you only use the meas_unit constraint (modified to draw its definition from the instrument definitions). Is that temporarily, i.e. why not the qubits resource (who else takes care of X q[0] and Y q[0] not being allowed in the same cycle?)? And the edge and detuned_qubits resources?

wvlothuizen:
this is still Work In Progress, but the qubit constraint is there (it takes it parameter from the main section). I need to discuss the edge/detuned_qubits resources with Miguel because adding them may interfere with the decomposition rules currently used as an alternative.

wvlothuizen:
dropping this PR in favour of #414

PR #399, opened 2021-03-24
Explicit definition of bundle start_cycle

wvlothuizen:
a less obtrusive cousin to PR #398


wvlothuizen:
dropped in favour of PR #410

PR #401, opened 2021-03-25
Multi core

jvansomeren:
Multi-core support for limiting bandwidth on communication channels between cores was added.
There are 2 versions: using topological special communication qubits or using a new channels resource in the rcscheduler.
Inter-core communication is supported by a new type of gate tagged "extern" (next to "mw", "flux", "readout", etc.).
All this is supported by new configuration file entries, with defaults set so that you don't use this multi-core extension.
Added tests.


jvanstraten:
I see a lot of changes that seem specific to CC-light? Is this multi-core thing actually something intended for CC-light somehow, or is it just an artifact of things not having been properly cleaned up yet?

jvansomeren:
Yes,the latter, to run the mapper, rcscheduler and use constraints a backend choice is still needed.
Instead of creating a dedicated backend, polluting the code base, I chose to extensie our reference and example backend with it. Codegen is not done.


Groet,

Hans van Someren



Op 25 mrt. 2021 om 17:37 heeft jvanstraten ***@***.***> het volgende geschreven:

﻿

I see a lot of changes that seem specific to CC-light? Is this multi-core thing actually something intended for CC-light somehow, or is it just an artifact of things not having been properly cleaned up yet?

jvanstraten (approved):

PR #414, opened 2021-05-28
Add CC support to resources

wvlothuizen:


src/ql/arch/cc/info.cc
```diff
@@ -587,7 +720,22 @@ void Info::post_process_platform(
  */
 void Info::populate_backend_passes(pmgr::Manager &manager, const utils::Str &variant) const {
 
-    // Mimic original CC backend.
+    // Remove passes we don't need/want.
+    // FIXME: bit of a hack, and invalidates https://openql.readthedocs.io/en/latest/gen/reference_architectures.html#default-pass-list
+    const auto &prescheduler = com::options::global["prescheduler"];
+    if (prescheduler.as_bool() && prescheduler.get_default() == prescheduler.as_str()) {    // prescheduler enabled implicitly
```
jvanstraten, line 726: I suppose I should have looked it up when this came up over e-mail. The correct way to check for this is `!prescheduler.is_set()`; is_set() returns true iff the option was explicitly configured by the user.

src/ql/arch/cc/info.cc
```diff
@@ -587,7 +720,22 @@ void Info::post_process_platform(
...
+        const utils::Str passes[] = {
+            "prescheduler",         // enabled by default, but pointless since we add our own scheduling
+            "scheduledqasmwriter"   // confusing, since it doesn't use our scheduling
```
jvanstraten, line 729: The scheduledqasmwriter pass should always be removed, since you always add it later in a later place. See also next comment.

src/ql/arch/cc/info.cc
```diff
@@ -599,6 +747,14 @@ void Info::populate_backend_passes(pmgr::Manager &manager, const utils::Str &var
...
         }
     );
+    manager.append_pass(
+        "io.cqasm.Report",
+        "scheduledqasmwriter",
+        {
+            {"output_prefix", com::options::global["output_dir"].as_str() + "/%N"},
+            {"output_suffix", "_scheduled.qasm"}
+        }
+    );
```
jvanstraten, line 757: I guess it would be better if there were an API to move existing passes, because then options would be preserved. That doesn't currently exist, but you can also preserve the options manually, rather than just substituting defaults again here. I would do something like:

```c++
if (manager.does_pass_exist("scheduledqasmwriter")) {
    auto options = manager.get_pass("scheduledqasmwriter").get_options();
    manager.remove_pass("scheduledqasmwriter");
    manager.append_pass("io.cqasm.Report", "scheduledqasmwriter").get_options().update_from(options);
}
```

I think that should handle everything properly, without duplicating the option defaults.

src/ql/arch/cc/resources/hwconf_default.json
```diff
@@ -829,42 +829,7 @@
         }
     },
     "topology": {
-        "edges": [
-            { "id":  0, "src":  2, "dst":  0 },
-            { "id":  1, "src":  2, "dst":  1 },
-            { "id":  2, "src":  2, "dst":  3 },
-            { "id":  3, "src":  2, "dst":  4 }
-        ]
     },
```
jvanstraten, line 832: It should be legal to just remove the whole topology key nowadays. Same for resources.

jvanstraten (changes requested):
Some notes about the way the "prescheduler" and "scheduledqasmwriter" are removed/shuffled around. Looks good otherwise.

wvlothuizen:
@jvanstraten : I addressed the review comments

jvanstraten (approved):

PR #416, opened 2021-05-31
Incremental changes for CC

jvanstraten:
Fixes a bug in the resource manager unrelated to CC, and adds a new hook to `arch::InfoBase` for doing stuff with the platform *after* the generic loading logic completes (complementing the hook for changing behavior *before* loading).


wvlothuizen (approved):
exactly fits my needs, thanks

PR #418, opened 2021-06-08
IR refactoring

jvanstraten:
This PR tracks the introduction of a new tree-gen-based IR structure for everything but the API. Some important new features:

 - Introduction of expressions and (builtin) functions, to track the functionality of cQASM.
 - Replacement of the hardcoded operand types to a generic operand list consisting of expressions.
 - Replacement of the hardcoded condition types with an expression.
 - Separation of instruction instances and types via links (implemented in tree-gen as `weak_ptr`s).
 - Removal of builtin gates aside from `wait` and the dummy source/sink nodes needed by the scheduler, except as needed for API compatibility.
 - Removal of ad-hoc decomposition, except as needed for API compatibility.
 - Separation of virtual and physical objects (aka variables and registers in the target), including qubits.
 - Probably more that I'm forgetting right now.

The old IR is completely retained for API compatibility; that is, the API functions will continue to use this structure. The old IR will however be completely converted to the new IR upon `program`/`compiler.compile()`. Therefore, all passes will need to be refactored.

TODO:

 - [x] Settle on a new IR structure.
 - [x] Initial refactoring to move the old IR out of the way of the new one. The old `ir` and `plat` namespaces are merged into `ir::compat`, allowing `ir` itself to be used for the new IR. `plat::Topology`'s new location is TBD; either `com` or `ir::prim`.
 - [x] Define the new IR using tree-gen.
 - [x] Build the conversion logic to go from the old IR to the new one.
 - [x] Build cQASM 1.2 writer logic.
 - [x] Build cQASM 1.2 reader logic.
 - [x] Build the conversion logic to go from the new IR back to the old one.
 - [x] Refactor pass management, using conversion logic before and after each "legacy" pass implementation.
 - [x] Data dependency graph construction & operations on it.
 - [x] Reimplement list scheduler.
 - [x] Add instruction decomposition/expansion logic.
 - [x] Add structured control-flow decomposition to basic-block form.
 - [x] Control-flow graph construction & operations on it.
 - [x] ~~(If there's time) add register allocation logic.~~

Not really related to the core of this PR, but related to wrapping up:

 - [x] Write docs on how you'd go about adding a pass or resource.
 - [x] Merge in diamond and make sure its annotations work with new-to-old and old-to-new.


jvansomeren (approved):

jvanstraten:
@jvansomeren Regarding the verbal comments:

 - The debug option not appearing to be completely documented was due to me looking at the online ReadTheDocs docs instead of my locally generated files.
 - The instruction decomposition logic was actually correctly recursively applied for structured control flow, it was just done in the pass rather than in the `com::dec` function. So I just updated the docstring of that function accordingly.

PR #425, opened 2021-07-23
Adopting CC for new IR

wvlothuizen:
This PR changes the CC backend by using the new IR instead of the old one. It now also supports features of the new IR like structured control flow.
While implementing, a few small changes were made to the core of OpenQL to fix some small issues and ease debugging. Additionally, some comments were added.


wvlothuizen:
Superseded by PR #433

PR #427, opened 2021-08-03
Fix assertion failure during DDG construction

jvanstraten:
Inconsistent event-node pair commutation rule application during DDG construction was causing assertion failures in some cases, specifically when a statement causes multiple events that would not commute with each other if they weren't due to the same statement.


wvlothuizen (approved):
Small change, large gain, thanks

PR #428, opened 2021-09-09
Fix statements like an infinite loop not causing any data dependency graph events

jvanstraten:
Constructs like

```
while (true) {
}
```

were resulting in invalid data dependency graphs, because they use no data whatsoever, therefore didn't get any dependency graph edges (not even to the source and sink sentinels), and thus becoming unschedulable islands. This adds a special case to the dependency graph builder to treat such statements as a barrier.

Note that while this generates a correct DDG and schedule and is usually what you want, this behavior is much more strict than what you might expect! For example, the following code would not hit the special case and thus would not become a barrier, allowing it to be scheduled anywhere in the program:

```
while (true) {
    set creg(0) = 0;
}
```

Therefore, be aware that if you're writing infinite loops to "terminate" the program, you need to put a barrier in front to ensure that all code you wrote before it is actually scheduled before it.

This PR also fixes a trivial null dereference bug for `for` loops without init or update statement and `while` loops that I'm pretty sure was already fixed in the CC branch.


wvlothuizen (approved):
Great, thanks again

jvansomeren:
Is this sufficient for all cases?
I was expecting a barrier before every label (join of control flow) and just before the real jump in every branch (split of control flow).

PR #433, opened 2022-03-07
new-IR support in CC backend

wvlothuizen:
This PR changes the CC backend by using the new IR instead of the old one. It now also supports features of the new IR like structured control flow.

While implementing, a few small changes were made to the core of OpenQL to fix some small issues and ease debugging. Additionally, some comments were added.

This PR supersedes #425, and adds much improved handling of cQASM functions to that


wvlothuizen:
Somehow Travis has come to life again, probably because it is enabled on QuTech-Delft. Although it doesn't seem to *always* fail, it does now on https://github.com/QE-Lab/eQASM_Assembler.git, which hasn't been a part of this project for quite some time.

Since the .travis.yml configuration file is outdated, and superseded by Github actions anyway, it will be deleted

include/ql/utils/json.h
```diff
@@ -17,7 +17,7 @@
 // check existence of JSON key within node, see PR #194
 #define QL_JSON_EXISTS(node, key)  ((node).count(key) > 0)
 
-#define QL_JSON_FATAL(s) QL_FATAL("Error in JSON definition: " << s)   // NB: FATAL prepends "Error : "
+#define QL_JSON_FATAL(s) QL_FATAL("Error in JSON definition: " << s)   // NB: FATAL prepends "Error : ". FIXME: replace by QL_JSON_ERROR
```
jvansomeren: I see many FIXMEs. Several, like this one, seem easy to implement.

src/ql/arch/cc/info.cc
```diff
@@ -178,7 +179,7 @@ void Info::dump_docs(std::ostream &os, const utils::Str &line_prefix) const {
                 {   "type": "mw",
                     "operand_idx": 0,
                     "value": [
-                        "{gateName}-{instrumentName}:{instrumentGroup}-gi",
+                        "{gateName}-{instrumentName}:{instrumentGroup}-gi",  // FIXME: outdated
```
jvansomeren, line 182: Also this FIXME.

src/ql/arch/cc/info.cc
```diff
@@ -447,10 +423,11 @@ void Info::dump_docs(std::ostream &os, const utils::Str &line_prefix) const {
        - triggering a measurement (on the UHFQA);
        - waiting for the internal processing time of the UHFQA;
        - retrieve the measurement result, and distribute it across the CC; and
-       - wait fot the data distribution to finish.
-
+       - wait for the data distribution to finish.
+)"
```
jvansomeren, line 427: Etc. Will now stop commenting on FIXMEs.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
```diff
@@ -479,63 +600,122 @@ void Codegen::bundleFinish(
...
+    [OPENQL] /Volumes/Data/shared/GIT/OpenQL/src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc:809 operand: q[10]
+
+    cQASM "cz q[0],q[10]" with JSON "cz q1,q10" results in:
+    [OPENQL] /Volumes/Data/shared/GIT/OpenQL/src/ql/arch/cc/pass/gen/vq1asm/detail/backend.cc:152 custom instruction: name=cz
+    [OPENQL] /Volumes/Data/shared/GIT/OpenQL/src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc:829 operand: q[0]
+    [OPENQL] /Volumes/Data/shared/GIT/OpenQL/src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc:829 operand: q[10]
+    */
+#if 1   // FIXME
+    if(!custom.instruction_type->template_operands.empty()) {
+        QL_DOUT("found template_operands: JSON = " << custom.instruction_type->data.data );
```
jvansomeren, line 646: The texts above with [OPENQL] were not clear to me. I agree that you insist on getting specialized instructions. Then that decomposition should have been done before codegen. Is that the message of those texts?

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
```diff
@@ -544,250 +724,432 @@ void Codegen::customGate(
...
              *             - no result, measurement results are often read offline from the readout device (mostly the raw values
              *             instead of the binary result), without the control device ever taking notice of the value
              *             - implicit bit result for qubit, e.g. for the CC-light using conditional gates
              *         - creg result (old, no longer valid)
              *             note that Creg's are managed through a class, whereas bregs are just numbers
              *         - breg result (new)
+             *
+             *  In the new IR (or, better said, in the new way "prototype"s for instruction operands van be defined
+             *  using access modes as described in
+             *  https://openql.readthedocs.io/en/latest/gen/reference_configuration.html#instructions-section
```
jvansomeren, line 770: Is this a new issue? Of what? Because it is the intention to support this in general. And then codegen could add constraints on the actual bit locations that get updated by a measurement.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
```diff
@@ -544,250 +724,432 @@ void Codegen::customGate(
...
-    RawPtr<const Json> pragma = settings.getPragma(iname);
-    if (pragma) {
-        for (Vec<BundleInfo> &vbi : bundleInfo) {
-            // FIXME: for now we just store the JSON of the pragma statement in bundleInfo[*][0]
-            if(vbi[0].pragma) {
-                QL_FATAL("Bundle contains more than one gate with 'pragma' key");    // FIXME: provide context
-            }
-            vbi[0].pragma = pragma;
+/************************************************************************\
+| Structured control flow
```
jvansomeren, line 819: OpenQL supports a 'structure decomposer' pass to decompose structured control flow to blocks and gotos.
The idea behind that is to have reusable code for this that need not be specialized in the codegen. What is the reason that this pass was not used?

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
```diff
@@ -544,250 +724,432 @@ void Codegen::customGate(
...
-/************************************************************************\
-| Some helpers to ease nice assembly formatting
-\************************************************************************/
+// FIXME: recursion?
+// FIXME: or pass SetInstruction or Expression depending on use
+// FIXME: adopt structure of cQASM's cqasm-v1-functions-gen.cpp register_into used for constant propagation
 
-// FIXME: assure space between fields!
-// FIXME: make comment output depend on verboseCode
+// FIXME: see expression_mapper.cc for inspiration
```
jvansomeren, line 987: Depending on the target's capabilities, there may be several walks over the expression here, even done by passes before this one, assuming that this codegen essentially only wants to do emits.
Examples:
when there is a mismatch between an expression node and a target function, then that expression node should be transformed to match the primitive target function first
allocation of intermediate registers, when needed, in general: register allocation
necessary conversion functions may have to be added
reading from a location may need address arithmetic to be done (which is also code)

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.h
```diff
@@ -25,123 +29,185 @@ namespace gen {
...
+     *
+     * A default set of functions that only handle constant arguments is provided by libqasm, see
+     * 'register_into(resolver::FunctionTable &table)'. These functions add a constant node to the IR when called (and
+     * fail if the arguments are not constant)
+     *
+     * Some of these are overridden by OpenQL to allow use of non-constant arguments. This is a 2 step process, where
+     * 'convert_old_to_new(const compat::PlatformRef &old)' adds functions to ir->platform using 'add_function_type',
+     * and 'ql::ir::cqasm:read()' then walks 'ir->platform->functions' and adds the functions using
+     * 'register_function()'. These functions add a 'cqv::Function' node to the IR (even if the arguments are constant,
+     * so overriding a function defeats libqasm's constant removal for that function).
```
jvansomeren, line 116: Is this a bug?

src/ql/arch/cc/info.cc
wvlothuizen: Removed the one just below these lines

src/ql/arch/cc/pass/gen/vq1asm/detail/operands.h
```diff
@@ -0,0 +1,149 @@
+/** \file
```
jvansomeren, line 1: Why is this not common code?

src/ql/arch/cc/resources/hwconf_default.json
```diff
@@ -248,682 +367,1323 @@
...
+        // flux
+        //**************************************************************************************************************
+
+        // helpers for CZ gate decomposition, flux_lutman compatible
+        //
+        // NB: the '*_park' instructions now use prototype "I:qubit" for the park parameter, which makes the scheduler
+        // *ignore* that parameter. This allows for parallel scheduling of multiple instructions that perform the same
+        // park, but can also arise in conflicts (signal conflict in the CC backend) if a qubit is both used for parking
+        // and another fluxing operation.
+
```
jvansomeren, line 1200: There used to be a resource in the old OpenQL to prevent this conflict. HvS presented an algorithm in our common meeting that could be used to generate these resource checks. Why is this comment then still here? This suggests that the implementation accepts that it every now and then fails.

include/ql/utils/json.h
wvlothuizen: Solved: there was actually only one remaining reference to the old QL_JSON_FATAL, now QL_JSON_ERROR (from exception.h)  is used throughout

src/ql/arch/cc/info.cc
wvlothuizen: The documentation is lagging behind a bit, will fix when things have stabilized

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
wvlothuizen: The texts with "[OPENQL]" are saved log messages for different combinations of operands being template operands or plain operands. For now, the existence of template operands is forbidden (in the backend), which means that gate decomposition must have been performed successfully.
This is still there because it does need work, especially to behave well on misconfiguration.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
wvlothuizen: I don't think it's an issue for now, it will become so if we want to do fancy feedback stuff.

It's a result of the available access modes, specifically "M", which states:
_M for a measurement of the qubit for which the result is written to its implicitly associated bit register (DDG W for the qubit and its bit, liveness use for the qubit, and liveness kill for the bit); and`_

Equivalent access modes that allow explicit specifications for a qubit and a bit separately do not seem to exist

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
wvlothuizen: I think there are 2 main reasons: 

- because the original program structure is lost, debugging becomes difficult
- specific optimisations become difficult. The Q1 processor has a 'loop' instruction that decrements a register and loops if the result is non-zero. The CC backend now uses that instruction when possible, doing so would be difficult if structure has been decomposed.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
wvlothuizen: Not sure what you mean. The current implementation does not try to generate code for every conceivable expression, but it does try to either work correctly, or give an appropriate error message. For one thing this seems to cover most of our needs, on the other hand doing things like a 'real' compiler is beyond the current scope of work.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.h
wvlothuizen: No (otherwise I would have added a "FIXME" ;-), it was designed this way, the comment is just there because it's not immediately apparent that things work this way (it took some emailing with Jeroen for me to grasp it). 

Note that https://github.com/QuTech-Delft/OpenQL/issues/424 contains ideas to make function handling more flexible (among many other things)

src/ql/arch/cc/pass/gen/vq1asm/detail/operands.h
wvlothuizen: two lines below there's this note:
`FIXME: could be useful for other backends and should be moved if this becomes appropriate
`
But making is common may require more generality than currently present.

There is also overlap with new_to_old.cc, but since that should eventually be deprecated it doesn't seem worth the effort.

src/ql/arch/cc/resources/hwconf_default.json
wvlothuizen: The experimentalists in the lab using OpenQL require full control over pulses (and scheduling), and therefore algorithms automating things aren't easily accepted, especially if they are sensed as a black box. Doing things manually - even if that sometimes results in a compiler error - is perceived as being easier (especially at this stage with 17 qubits only), and it allows exploring new ideas, such as a CZ that not only parks some qubit, but also performs 'refocussing', flux assisted measurement, etc.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
jvansomeren: This issue is not urgent, merely to reduce later development time, e.g. when one of the two structure decomposition methods is to be synchronized with the other one because of certain benefits. So, I'm exploring now the reasons behind so that we learn from this.

Keeping track on correspondence between input and output program, can be done through code location annotations.

Yes, the structural decomposition pass is all or nothing. In the case of CC with static loop support, one would want lowering of if-then-else, repeat-until, while-do, dynamic-for but not static loop support. This could be established by adding options to the structural decomposition pass to selectively enable the respective individual lowerings. Would you consider this an option?

Reducing code duplication saves debugging time and future development time (you know of course).
For instance dealing with the following: nested static loops, updates to the IR, adding if-convert (turning cheap if-then-elses to conditionally executed code).

So, your opinion please.

wvlothuizen: Options to the structural decomposition pass may be of help, although I imagine it may be cleaner to rewrite the decomposition pass according to the lines of https://tree-gen.readthedocs.io/en/latest/interpreter.gen.html#visitor-pattern, and then override from the decomposition class in the backend wherever the default isn't suitable.

Then again the 'problem' is far worse for expressions than it is for structured control flow.

https://github.com/QuTech-Delft/OpenQL/issues/424 contains quite a few good ideas to handle commonalities in a better fashion

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
jvansomeren: The implicit association bothers me. Requiring implementing this for each target is wrong. I know there is some backward compatibility here and also cQASM compatibility stuff. But those latter can be get rid off. I would like to take a step back here and investigate without taking care of those backward and cQASM compatibility issues.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
jvansomeren: Separating walking down the tree and visiting the nodes, allowing overriding the latter, is a better way to do this, indeed.
We need to make a real step forward here.
Next to the visitor pattern, we need a transform pattern that works similarly but returns a replacement IR fragment at each level.
And in both we need a context parameter (kind of stack) to do context dependent visits/transformations.
Furthermore, the IR may contain structures that are not trees; the related classes should also be extended with visitor-like methods; those then maintain the sequencing through the non-tree. Similarly, for trees there are several walks possible rather then the depth-first, left-to-right one. I didn't look but perhaps the visitor pattern can be instantiated for each of these. E.g. each node maybe visited twice, before recursing and after recursing down its children.
This what concerns the visitor pattern.

Code generation for statements and expression trees is indeed far from trivial. Look at all those 'code generators' around. Please indicate when you think you need to take a step back and add more support for this. For one thing, I learned that it is easier to make multiple traversals, gradually annotating and transforming a tree rather than doing everything in one go. It may be illustrative when I once give a high-level overview of code generation in CoSy. CoSy has passes, visitor/transform patterns, generated pattern matcher to transform to target code based on configuration, storage allocation passes, pass plug-ins, etc.  CoSy lets itself be read in a top-down form so perfect for learning.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
wvlothuizen: Many things are 'wrong' with the way "measure" is currently defined. First of all, the implicit result bit, which was in the original language spec, also see https://libqasm.readthedocs.io/en/latest/cq1-instructions.html#default-instruction-set. On the one hand it does not properly allow assigning to some explicit bit (one would like to be able to write `b[23] = measure(q[0])`, on the other hand it does not allow qubits (that are part of the qubit register) to be measured without storing the bit result.
Then there is the assumption that a measurement produces binary data, and not e.g. 1 of 3 possibilities. And then we have calibration, where measurements are performed that produce (e.g.) raw IQ values.

My proposal would be to create an issue for this, as it is out of scope of the current PR

src/ql/arch/cc/resources/hwconf_default.json
jvansomeren: I appreciate that the experimentalists want to be in full control and need support for that. That is the idea behind making passes visible and all functionality transparent: that 'compiler-intelligence' can always be overridden.
But the code generator passes are kind of inevitable: every compiler for transmon needs to run them.
The experimentalists are not the only ones using transmon and its compiler.
For one, there are the QuantumInspire users that like to use a real machine for some reason.
For them, there should be a more fool-proof but also more open, research-encouraging compiler implementation that doesn't run into errors on some circuits.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
jvansomeren: Agree.

src/ql/arch/cc/pass/gen/vq1asm/detail/codegen.cc
jvansomeren: In parallel to your work, we are rewriting old-IR passes to new-IR passes, e.g. clifford and later mapper.
In those contexts I will try to make this step forward as well.

wvlothuizen: I think it would be good to keep an open eye on commonalities when more passes are migrated to the new IR. The CC was the first to make the switch I guess, and it leans heavily on new_to_old. Since that is to be removed in the end, I didn't particular care about some copy/paste/edit steps that I wouldn't normally find acceptable.

src/ql/arch/cc/resources/hwconf_default.json
wvlothuizen: For QI users is may indeed we fruitful to create a separate pass for qubit parking, especially when the number of qubits becomes larger that the current 5 (where conflicts are sort of impossible).
Still I do think it's inevitable that users can construct syntactically sound circuits that still cannot run on some actual hardware, and thus results in an error. The way user/input errors are reported in OpenQL should definitely be improved.

src/ql/com/ddg/build.cc
```diff
@@ -427,7 +422,12 @@ class Builder {
         }
         if (!any_edge) {
             for (const auto &nc : global_write) {
-                QL_ASSERT(!nc.get().commutes_with(incoming));
+                if(nc.get().commutes_with(incoming)) {
```
jvansomeren, line 425: Is this there because sometimes the depgraph is circular?
Then that should be an issue. I still have to review depgraph construction, especially in the context of commuting gates. I suspect an error in the latter handling. Have you ever run a compiler with one of those commuting options on?

wvlothuizen: No, during early testing the original QL_ASSERT triggered, but for other reasons than circularity. I replaced the assert by a somewhat more informational message. Jeroen solved the underlying problem, see https://github.com/QuTech-Delft/OpenQL/pull/427 and https://github.com/QuTech-Delft/OpenQL/pull/428 for background

src/ql/ir/compat/detail/cqasm_reader.cc
```diff
@@ -603,7 +603,7 @@ lqa::Analyzer ReaderImpl::build_analyzer() {
     a.register_function("operator!=", "bb", op_lxor_bb);
     for (const auto &gate : gateset) {
         a.register_instruction(gate->cq_insn);
-        a.register_instruction("skip", "i", false, false);
+        a.register_instruction("skip", "i", false, false);  // FIXME: why inside loop
```
jvansomeren: I propose to take it out of the loop, i.e. put it after the loop then. I also see no reason why it is in the loop, especially after analyzing libqasm on this.

src/ql/ir/new_to_old.cc
```diff
@@ -486,10 +486,24 @@ void NewToOldConverter::convert_block(
                     // Handle the normal operands for custom instructions.
                     Operands ops;
                     for (const auto &ob : custom->instruction_type->template_operands) {
-                        ops.append(*this, ob);
+                        try {
+                            ops.append(*this, ob);
+                        } catch (utils::Exception &e) {
+                            e.add_context("name="+custom->instruction_type->name+", qubits="+ops.qubits.to_string());
```
jvansomeren, line 492: Shouldn't this be in append instead of here? Then appends needs knowledge of the instruction_time.
There are many more ops.append's where this context is not added.

src/ql/ir/new_to_old.cc
```diff
@@ -486,10 +486,24 @@ void NewToOldConverter::convert_block(
...
+                            throw;
+                        }
                     }
                     for (utils::UInt i = 0; i < custom->operands.size() - diamond_op_count; i++) {
-                        ops.append(*this, custom->operands[i]);
+                        try {
+                            ops.append(*this, custom->operands[i]);
+                        } catch (utils::Exception &e) {
+                            e.add_context(
+                                "name=" + custom->instruction_type->name
```
jvansomeren, line 501: Same as above.

src/ql/ir/old_to_new.cc
```diff
@@ -229,14 +231,16 @@ static void parse_decomposition_rule(
             "expansions"
```
jvansomeren, line 231: At line 63 above, there is \"breg\", it should be \"b\", since the substr(0,1) higher up results in a string of 1 long.
Can you make this patch in your branch?

src/ql/ir/old_to_new.cc
```diff
@@ -229,14 +231,16 @@ static void parse_decomposition_rule(
...
     }
-    utils::UInt decomp_duration = get_duration_of_block(rule_ir->program->blocks[0]);
-    if (decomp_duration > ityp->duration) {
-        QL_USER_ERROR(
-            "in " << description.str() << ": the duration of the schedule of " <<
-            "the decomposition (" << decomp_duration << ") cannot be longer " <<
-            "than the duration of the to-be-decomposed instruction (" <<
-            ityp->duration << ")"
-        );
+    if (ityp->duration > 0) {   // 0 implies that durations are just added up, requires scheduling after decomposition
```
jvansomeren, line 234: I don't understand this.
Is allowing a duration of 0 a new convention for instruction definitions in the config with new-style-decompositions? I see that without this, you would get an error when the decomposition contains instructions with >0 duration. So, when you know that you would schedule anyhow, you don't want this error.
So, setting the duration to 0, avoids this error, nothing else.
What you would want to define with a particular decomposition is, that it must always be scheduled first, i.e. when the error condition applies, the cycle assignments in the block in which it gets expanded, become invalid.
So we are back at the cycles_invalid flag.

src/ql/ir/old_to_new.cc
```diff
@@ -460,17 +467,27 @@ Ref convert_old_to_new(const compat::PlatformRef &old) {
 
                 // Check whether the operand list matches the specializations
                 // specified in the old way.
-                if (insn->template_operands.size() > insn->operand_types.size()) {
+                // FIXME: debug WIP, also see use of template_operands further down
```
jvansomeren: Can you explain more on this? What is the problem? I guess you have a use case!

src/ql/ir/old_to_new.cc
```diff
@@ -866,17 +884,21 @@ Ref convert_old_to_new(const compat::PlatformRef &old) {
         }
     }
 
+#if 0    // FIXME: moved down, see comment overthere
```
jvansomeren: You clearly have struggled here. See request above.

src/ql/ir/compat/detail/cqasm_reader.cc
wvlothuizen: Fixed in commit 0cb35fbd

src/ql/ir/old_to_new.cc
wvlothuizen: performed in commit f5a78ac0

Note that you can attach a review comment to a specific line of code

src/ql/ir/old_to_new.cc
wvlothuizen: Your analysis is correct, this allows disabling the check, allowing the decomposition to take place. It doesn't make sense to schedule an instruction with 0 duration, and decompose it afterwards.
The new IR - like the old one - can represent both unscheduled and scheduled programs. I don't know whether - and if so, where - this state is maintained or checked.

src/ql/ir/old_to_new.cc
wvlothuizen: I'll clean this up, it was added while debugging

src/ql/ir/old_to_new.cc
wvlothuizen: There was a bug which made `check_consistency()` fail if decompositions contained cQasm, because at this original location `topology` hasn't been populated yet.
I'll cleanup this one too.

wvlothuizen: Solved in commit cd5efefa

src/ql/ir/old_to_new.cc
wvlothuizen: solved in commit 17d7625c

src/ql/ir/old_to_new.cc
jvansomeren: I'm exploring how to address this, since setting the instruction duration to 0 is not general enough in the context of multiple decompositions, for different mapping strategies, for instance.
The documentation reads:

> the duration of the decomposed block may not be longer than the duration of the to-be-decomposed instruction (either make the instruction duration long enough, or define the decomposition as a single bundle and (re)schedule after applying the decomposition).

When you certainly will schedule after decomposition, you can as well specify the instruction's duration to be larger than you ever think the decomposition will take. I'd prefer that approach. It is also in line with the documentation.

To have a system of predicates around passes such as: `circuit-cycles-valid` is quite advanced, when you want to make it a general feature; other examples are: `has-been-decomposed-to-primitives` and `is-mapped-to-topology` which of course depend on the configuration/target.

src/ql/ir/ops.cc
```diff
@@ -343,6 +343,7 @@ InstructionTypeLink find_instruction_type(
 
     // If we shouldn't generate an overload if only the name matches, stop now.
```
jvansomeren, line 344: Just a note: I don't like the use of the term overload for an instruction type / prototype.
An overload would mean to me a companion instruction (or set of those) that also matches the instruction type, those overload each other.
Do you also see this?
Then I'd like to update the documentation/sources in this direction, once this review has completed.

src/ql/pmgr/pass_types/base.cc
```diff
@@ -1000,15 +1000,20 @@ void Base::handle_debugging(
...
-        ir->dump_seq(
-            utils::OutFile(context.output_prefix + "_debug_" + in_or_out + ".ir").unwrap()
-        );
-        ir::cqasm::WriteOptions write_options;
-        write_options.include_statistics = true;
-        ir::cqasm::write(
-            ir, write_options,
-            utils::OutFile(context.output_prefix + "_debug_" + in_or_out + ".cq").unwrap()
-        );
+        if (!after_pass && ir->program.empty()) {  // don't dump if we haven't got a program yet
```
jvansomeren: Why this check on after_pass?
I can imagine that the compiler ends with a qasm_writer and then runs a pass that cleans the IR (because it after that wants to read something else into the IR) and in between the program is empty. So I propose to omit the after_pass check and keep the program is empty check.

src/ql/ir/ops.cc
wvlothuizen: I'd say that the term 'overload' is used correctly here: you can have several instruction definitions with the same name, but different prototypes.

src/ql/pmgr/pass_types/base.cc
wvlothuizen: Originally, the dumping was unconditional, which didn't make sense before running the very first pass, since the IR tends to be empty then (at least if that pass is `io.cqasm.read`). 
The condition introduced here was meant to be the most restrictive and thus give the least interference. It may be noteworthy to someone that is debugging if the IR becomes empty along the way.

src/ql/pmgr/pass_types/base.cc
```diff
@@ -1120,68 +1125,79 @@ void Base::compile(
...
+            default: QL_ASSERT(false);
+        }
 
-    // Handle configured debugging actions after running the pass.
-    handle_debugging(ir, context, true);
+        // Handle configured debugging actions after running the pass.
+        compile_phase = "debugging.after";
+        handle_debugging(ir, context, true);
 
+    } catch (utils::Exception &e) {
```
jvansomeren, line 1191: Where do the exceptions that you want to catch come from? Only from handle_debugging or from executing the passes?
Or, why is this compile_phase relevant?
Or, why did you add this exception handler?

wvlothuizen: Exceptions typically originate from QL_ICE and friends. This exception handler adds the pass name and 'phase' as context, and then rethrows. It was added because it was typically difficult to pinpoint where and when things went wrong, especially if an error occurred in code that's common between passes (such as `handle_debugging()`)

tests/cc/test_cond_gate.cq
```diff
@@ -0,0 +1,96 @@
...
+
+cond (b[1]) x q[0]
+cond (!b[1]) x q[0]
+
+cond (b[0] && b[1]) x q[0]
+cond (b[0] || b[1]) x q[0]
+cond (b[0] ^^ b[1]) x q[0]
+
+# FIXME: seems to break CI, see https://github.com/QuTech-Delft/OpenQL/runs/5588352276?check_suite_focus=true
+# cond (rnd(0.5)) x q[0]
```
jvansomeren, line 42: What does rnd(0.5) do, rounding or random?
Failing on random is ok because it is not predictable.

tests/cc/test_cond_gate.cq
```diff
@@ -0,0 +1,96 @@
...
+
+# from: deps/libqasm/src/cqasm/tests/v1-parsing/semantic/insn-condition-new-ok/input.cq
+# note; differing lengths between condition and parameters are allowed
+# cond (b[0:3]) x q[4:5] # Error: Unknown error: in pass reader, phase main: Duplicate node of type N2ql2ir10ExpressionEat address 0x7f9e8b5d9978 found in tree
```
jvansomeren, line 46: Is this an issue?

tests/cc/test_looping.cq
```diff
@@ -0,0 +1,222 @@
...
+foreach (i = 10..0) {
+}
+
+
+.foreachLoopDown
+foreach (i = 10..5) {
+}
+
+
+# FIXME: we do not handle negative numbers
```
jvansomeren: Who is 'we', OpenQL or CC?

tests/cc/test_looping.cq
```diff
@@ -0,0 +1,222 @@
...
+#.condBit
+#while (mAncilla3) { # Error: encountered incompatible object reference to mAncilla3 (size=0)
+#}
+
+.condBit2
+while (b[0]) {
+}
+
+#.condBit3
+#while (b[0:2]) { # Error: Usage error: in pass reader, phase main: unexpected number of single-gate-multiple-qubit bit indices specified; found 3, expected 1
```
jvansomeren, line 172: See above, so sgmq handling in pass reader is not right.
What is the problem?

tests/cc/test_looping.cq
```diff
@@ -0,0 +1,222 @@
...
+
+
+
+
+
+# FIXME: move to its own file
+.more
+# set i = true # Error: type of right-hand side (bool/bit) could not be coerced to left-hand side (int reference)
+# set i = 4<5
+# set i = int(1+4) # Error: Input error: in pass VQ1Asm, phase main: in block 'more': in expression '1 + 4': cannot handle parameter combination '1' , '4'. FIXME: no constant propagation
```
jvansomeren, line 215: This one and those below, these seem to be errors that need solving. Is this in issue on the relation between constant propagation and function definitions? I assume that constant propagation and evaluation is done recursively until a single literal when all functions are known.

src/ql/arch/cc/resources/hwconf_default.json
jvansomeren: Can you make an issue that a compile setup (with new passes or with existing passes with new options) should be created that is suitable for QI users, i.e. that don't crash on a wrong qubit parking setup?
All circuits that users can write and that cannot be run on some actual hardware must get a decent error message; this requires adding passes that check this and then fail with a clear message.

src/ql/ir/old_to_new.cc
jvansomeren: So I propose that you give the instruction a long duration and take the test on the duration being 0 out.

tests/cc/test_cond_gate.cq
wvlothuizen: This is for future testing, and is likely to change

tests/cc/test_cond_gate.cq
wvlothuizen: I'm not sure

tests/cc/test_looping.cq
wvlothuizen: CC, updated in commit 73b3b2c7

tests/cc/test_looping.cq
wvlothuizen: Not sure, see above

src/ql/pmgr/pass_types/base.cc
jvansomeren: Yes, but then it is sufficient that the test is just` ir->program.empty()`.
I.e. omit `!after_pass,` because `after_pass` is also a reason when the program is empty to enter the if.

tests/cc/test_cond_gate.cq
jvansomeren: My intuition is that the meaning of this is that the conditional expression is the AND of the bits,
and that the q[4:5] is sgmq notation.
So these are separate and different uses of index ranges which need to be handled at different places in the reader.
So my intuition says that this is a issue of the reader.
To verify this, we need to review the libqasm cqasm1.2 definition and the reader's code.
If it is an error, an issue must be created in order not to forget it.

tests/cc/test_looping.cq
wvlothuizen: line 213 fails rightfully, it requires an `int()` cast
line 214 & 215 fail, because the backend does not support functions on constants (`operator<(4,5)`), since that is better done in a separate pass (working on that). Libqasm's constant propagation would handle it, but only works on functions that are not overridden by OpenQL.

src/ql/ir/old_to_new.cc
wvlothuizen: With the 'predicates around passes', I guess you mean to somehow maintain in the IR in what state it is, and then act upon that. 
I'm not so sure whether it's possible to come up with a generic solution for that: a scheduler may set `circuit-cycles-valid`, but then again a later pass might invalidate that. So the semantics of these predicates would have to be globally agreed upon, and new passes will probably need new predicates, requiring updates on all other passes. 

Another approach, which is more or less the current approach, would be for every pass to check the IR on it's requirements, and fail if these are not met (e.g. if the CC backend encounters a non-native instruction that should be decomposed).

Anyway, this would be a discussion for future development, a bit out of scope for this PR

src/ql/arch/cc/resources/hwconf_default.json
wvlothuizen: done: https://github.com/QuTech-Delft/OpenQL/issues/445

src/ql/pmgr/pass_types/base.cc
wvlothuizen: I still think that the more restrictive the condition for suppression is, the better. Your proposal would also suppress debug output after a pass runs, if the program is empty, while there might (generally speaking) be interesting data in `ir->platform` (not with the existing passes though)

src/ql/ir/old_to_new.cc
jvansomeren: My discussion of predicates around passes was more a reduction ad absurdum to come to the same conclusion as you do, that it is too far fetched now.
But I don't like the test on the duration being 0, or I don't understand the context of the current code, i.e. where this 0 duration comes from, from the config file or from somewhere internally in this old_to_new code.
I'm now trying to understand this.

tests/cc/test_cond_gate.cq
wvlothuizen: Just checked the reference from line 44 (`deps/libqasm/src/cqasm/tests/v1-parsing/semantic/insn-condition-new-ok/input.cq`), which is:
```
version 1.0
qubits 10

# note; differening lengths between condition and parameters are allowed
cond (b[0:3]) x q[4:5]
```

So this indeed seems to be a bug in cQASM reader

wvlothuizen: filed https://github.com/QuTech-Delft/OpenQL/issues/446

src/ql/pmgr/pass_types/base.cc
jvansomeren: By all this reasoning, is it then not better to also dump when the program is empty?
There may be interesting data in ir->platform upon entering the pass by the same argument.
Or do I miss something here?

src/ql/ir/ops.cc
jvansomeren: I'm still studying on the relations between all of the following:

- overload and instruction type
- generalization and specialization
- which legacy decomposition rules apply
- prototypes and operand types

in the context of old to new conversion of a platform configuration file.
See also some of the other conversations.

tests/cc/test_looping.cq
jvansomeren: I suggest to add to investigate this in the same issue as just created (on cond(b[...]) x q[...]). It is similar.

src/ql/ir/old_to_new.cc
jvansomeren: This is about adding legacy decompositions to instruction types in the new IR. Those legacy decomposition bodies (I'm guessing now) are resolved here (the instruction types of the sub instructions are found) when adding those to the to-be-decomposed instruction type.
I would have expected the error that the decomposition takes more cycles than the original instruction, to be given by the decomposition pass when actually performing the decomposition, and not here when adding this decomposition to the instruction type. When done by the decomposition pass, it could use the pass value ignore_schedule in giving this error message, because when the schedule is ignored doing this decomposition, the expansion can as well take more cycles than the original.

tests/cc/test_looping.cq
wvlothuizen: Added it to https://github.com/QuTech-Delft/OpenQL/issues/446

src/ql/ir/ops.cc
jvansomeren: This is not a comment on this PR.

src/ql/pmgr/pass_types/base.cc
jvansomeren: Or, why did you add this check? Because OpenQL ran into an error because the program is empty? Then that should be solved because dumping an empty program should be possible. Or because it is kind of non-informative? Then I prefer to leave this in.

src/ql/ir/old_to_new.cc
jvansomeren: Why did you add this test? A crash or convenience or was it a bug that it was absent?

jvansomeren: Perhaps it is easier to short-circuit these written discussions by discussing them in a meeting.

wvlothuizen: The test is part of function
```
/**
 * Parses a new-style decomposition rule.
 */
static void parse_decomposition_rule(
    const Ref &ir,
    const InstructionTypeLink &ityp,
    const utils::Json &json
) {
```
, a static helper for the *new-style* decompositions.

`ityp->duration` is taken from the `duration` field of the instruction in the JSON platform file.

When decompositions are intended to be used pre-scheduling, it is quite inconvenient to be forced to specify some duration. Specifying some 'large' number would work, but the trouble I have with that is that you then get into the situation where - depending on context - numbers are meaningful or not, and the context may be difficult to tell. A value that *explicitly* triggers special treatment is much more future-proof in my opinion, and 0 seems to be a safe value.

jvansomeren:
What was your use case?
Why did you add this test?

So for you a sufficiently high value would work as well.
Or we take this test here out and put it at those places where the decomposition is actually done.
And use the ignore_schedule flag.

That an instruction can have a duration and a decomposition serves the purpose to treat it as a primitive until the decomposition is done.
E.g. scheduled, optimized, and mapped.

Groet,

Hans van Someren



Op 13 apr. 2022 om 22:45 heeft Wouter Vlothuizen ***@***.***> het volgende geschreven:

﻿

@wvlothuizen commented on this pull request.

In src/ql/ir/old_to_new.cc<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_QuTech-2DDelft_OpenQL_pull_433-23discussion-5Fr849885057&d=DwMCaQ&c=XYzUhXBD2cD-CornpT4QE19xOJBbRy-TBPLK0X9U2o8&r=kNdT9ewT6pQdYFkBLR_5-ZqsrSTk7k5Hdd7MSC_Vnzg&m=m0y0dhwAHwZuMz4pmwiG4Vt2xNV1PH-iJVHEZ9YHJ4w&s=G4T43yFxVgOXQdO7JDwQ1dgBxorBpa-kLMkUIY6W5RA&e=>:

> @@ -229,14 +231,16 @@ static void parse_decomposition_rule(
             "expansions"
         );
     }
-    utils::UInt decomp_duration = get_duration_of_block(rule_ir->program->blocks[0]);
-    if (decomp_duration > ityp->duration) {
-        QL_USER_ERROR(
-            "in " << description.str() << ": the duration of the schedule of " <<
-            "the decomposition (" << decomp_duration << ") cannot be longer " <<
-            "than the duration of the to-be-decomposed instruction (" <<
-            ityp->duration << ")"
-        );
+    if (ityp->duration > 0) {   // 0 implies that durations are just added up, requires scheduling after decomposition


The test is part of function

/**
 * Parses a new-style decomposition rule.
 */
static void parse_decomposition_rule(
    const Ref &ir,
    const InstructionTypeLink &ityp,
    const utils::Json &json
) {


, a static helper for the new-style decompositions.

ityp->duration is taken from the duration field of the instruction in the JSON platform file.

When decompositions are intended to be used pre-scheduling, it is quite inconvenient to be forced to specify some duration. Specifying some 'large' number would work, but the trouble I have with that is that you then get into the situation where - depending on context - numbers are meaningful or not, and the context may be difficult to tell. A value that explicitly triggers special treatment is much more future-proof in my opinion, and 0 seems to be a safe value.

src/ql/ir/old_to_new.cc
wvlothuizen: *Comment by Hans that appeared lower down because Github's email interface messes up things*

What was your use case?
Why did you add this test?

So for you a sufficiently high value would work as well.
Or we take this test here out and put it at those places where the decomposition is actually done.
And use the ignore_schedule flag.

That an instruction can have a duration and a decomposition serves the purpose to treat it as a primitive until the decomposition is done.
E.g. scheduled, optimized, and mapped.

wvlothuizen: The use case is basically any new style decomposition, but a maybe interesting example is this:

```
        "_do_measure": {
            "prototype": ["M:qubit"],
            "duration": 800,
            "cc": {
                "signal": [
                    {   "type": "measure",
                        "operand_idx": 0,
                        "value": ["dummy"] 
                    }
                ],
                "static_codeword_override": [0] 
            }
        },
        "measure": {
            "prototype": ["M:qubit"],
            "duration": 0,
            "decomposition": {
                "into": "_do_measure op(0)"
            }
        },
        "measure q6": {
            "prototype": ["M:qubit"],
            "duration": 0,
            "decomposition": {
                "into": [
                    "rx12 q[6];",
                    "_do_measure q[6]"
                ]
            }
        },
```

The purpose here is to treat measurements of some qubits (q6 in this case) differently, by prepending a gate (`rx12`, a rotation on the second excited state) to improve measurement fidelity. Note that these decompositions (need to) happen before scheduling.

The actual measurement ("_do_measure") has a duration (800) that is actually generated by a script where the user can configure durations of particular types of gates, so doing the math by hand (800 for "measure" and 820 for "measure q6") is impractical and error-prone. The two decompositions of "measure" use the change discussed here to allow a duration of 0 to mean 'ignore'.

Of course is would be *possible* to use a large number (1000?, 1000000?), but these numbers tend to come with assumptions that may not hold over time, so I think it's bad design to require that. And I don't think that use of the value 0 for special treatment could ever conflict with the wish for a decomposition to actually take zero time.

But I guess we just disagree on what's the best solution

jvansomeren:
I'm still not confident with setting the instruction's duration to 0 to avoid getting the error.
See my remark on that e.g. the measure is seen as a primitive before the decomposition is done and then needs a duration with a value that is approximately its real value, so that optimizers can judge what is more optimal; a duration of 0 means e.g. that in scheduling the chain of instructions with the measurement is not critical. Note that 'scheduling' is not rcscheduling but using the depgraph as a representation of the circuit which takes durations and dependences into account, as e.g. the mapper does.
So getting the error is ok, not getting the error when the duration is 0, is also ok, but the means to not get the error by setting the duration to 0 is not ok for me.
I'm looking for a solution in which the duration can be approximately 800 and you still wouldn't get the error because the decomposition is done before scheduling. I'm inclined to move the test to where the decomposition is really done.
Also the test now is not sufficient because the decomposition may still lead to a failure when the resource use pattern of the substituted result is not matching that of the key gate.

src/ql/pmgr/pass_types/base.cc
wvlothuizen: Just checked, and without the extra test an exception is raised, basically telling that the Program is empty:

```
E       RuntimeError: Unknown error: in pass reader, phase debugging.before: dereferencing empty Maybe/One object of type N2ql2ir7ProgramE
E       Stack trace (most recent call last):
E       #31   Object "python3.6", at 0x107d5a25f, in _PyFunction_FastCallDict + 575
E       #30   Object "python3.6", at 0x107d596c2, in _PyEval_EvalCodeWithName + 2866
E       #29   Object "python3.6", at 0x107d55c0a, in _PyEval_EvalFrameDefault + 25626
E       #28   Object "python3.6", at 0x107c7aa44, in PyObject_Call + 100
E       #27   Object "python3.6", at 0x107c7ad82, in _PyObject_Call_Prepend + 130
E       #26   Object "python3.6", at 0x107c7ac05, in _PyObject_FastCallDict + 181
E       #25   Object "python3.6", at 0x107d5a25f, in _PyFunction_FastCallDict + 575
E       #24   Object "python3.6", at 0x107d596c2, in _PyEval_EvalCodeWithName + 2866
E       #23   Object "python3.6", at 0x107d55a11, in _PyEval_EvalFrameDefault + 25121
E       #22   Object "python3.6", at 0x107d58ab8, in call_function + 536
E       #21   Object "python3.6", at 0x107d59f38, in fast_function + 552
E       #20   Object "python3.6", at 0x107d55a11, in _PyEval_EvalFrameDefault + 25121
E       #19   Object "python3.6", at 0x107d58ab8, in call_function + 536
E       #18   Object "python3.6", at 0x107d59f38, in fast_function + 552
E       #17   Object "python3.6", at 0x107d55a11, in _PyEval_EvalFrameDefault + 25121
E       #16   Object "python3.6", at 0x107d58ab8, in call_function + 536
E       #15   Object "python3.6", at 0x107d59f38, in fast_function + 552
E       #14   Object "python3.6", at 0x107d55a11, in _PyEval_EvalFrameDefault + 25121
E       #13   Object "python3.6", at 0x107d58ae7, in call_function + 583
E       #12   Object "python3.6", at 0x107cc83d5, in _PyCFunction_FastCallDict + 165
E       #11   Object "_openql.cpython-36m-darwin.so", at 0x10a65e6af, in _wrap_Compiler_compile_with_frontend(_object*, _object*) + 159
E       #10   Object "_openql.cpython-36m-darwin.so", at 0x10ab68541, in ql::api::Compiler::compile_with_frontend(ql::api::Platform const&) + 49
E       #9    Object "_openql.cpython-36m-darwin.so", at 0x10a9b4dec, in ql::pmgr::Manager::compile(ql::utils::tree::base::One<ql::ir::Root> const&) + 108
E       #8    Object "_openql.cpython-36m-darwin.so", at 0x10a9a0dfd, in ql::pmgr::pass_types::Base::compile(ql::utils::tree::base::One<ql::ir::Root> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 2749
E       #7    Object "_openql.cpython-36m-darwin.so", at 0x10a9a02f6, in ql::pmgr::pass_types::Base::run_sub_passes(ql::utils::tree::base::One<ql::ir::Root> const&, ql::pmgr::pass_types::Context const&) const + 150
E       #6    Object "_openql.cpython-36m-darwin.so", at 0x10a9a0932, in ql::pmgr::pass_types::Base::compile(ql::utils::tree::base::One<ql::ir::Root> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 1522
E       #5    Object "_openql.cpython-36m-darwin.so", at 0x10a99f4c8, in ql::pmgr::pass_types::Base::handle_debugging(ql::utils::tree::base::One<ql::ir::Root> const&, ql::pmgr::pass_types::Context const&, bool) + 1176
E       #4    Object "_openql.cpython-36m-darwin.so", at 0x10a87f334, in ql::ir::cqasm::write(ql::utils::tree::base::One<ql::ir::Root> const&, ql::ir::cqasm::WriteOptions const&, std::__1::basic_ostream<char, std::__1::char_traits<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 84
E       #3    Object "_openql.cpython-36m-darwin.so", at 0x10a87f3bb, in ql::ir::cqasm::write(ql::utils::tree::base::One<ql::ir::Root> const&, ql::utils::tree::base::One<ql::ir::Node> const&, ql::ir::cqasm::WriteOptions const&, std::__1::basic_ostream<char, std::__1::char_traits<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 43
E       #2    Object "_openql.cpython-36m-darwin.so", at 0x10a887bfc, in ql::ir::cqasm::Writer::visit_root(ql::ir::Root&) + 204
E       #1    Object "_openql.cpython-36m-darwin.so", at 0x10a8366ef, in ql::utils::tree::base::Maybe<ql::ir::Program>::deref() const + 255
E       #0    Object "_openql.cpython-36m-darwin.so", at 0x10a6b9f71, in ql::utils::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ql::utils::ExceptionType) + 273

pybuild/openql/openql.py:2088: RuntimeError
```

jvansomeren: In the root node of the IR, program is a Maybe field.
That implies that every access to it must check whether it is empty, because that is an allowed condition.
src/ql/ir/cqasm/write.cc visit_root line 219 doesn't do that.
It just starts to visit the Program node.
So the bug is in visit_root, it should check emptiness of Program and if so, just skip visiting the program.

As a generalization, we could check that all maybe etc. nodes, when accessed, are checked for their emptiness in the code.

wvlothuizen: you're right, I will change write.cc, and revert the change here (next week, when I'm back from the US)

jvansomeren (commented):
No further review comment.
Why are so many tests failing?

wvlothuizen:
@jvansomeren test now pass

jvansomeren (approved):
Thanks, everything seems ok.
A long journey but I'm glad it can me merged in now
and the group may start using the new functionality.

PR #434, opened 2022-03-09
old scheduler criticality recursion taken out

jvansomeren:


src/ql/ir/compat/kernel.cc
```diff
@@ -553,8 +553,10 @@ Bool Kernel::add_custom_gate_if_available(
 
     // first check if a specialized custom gate is available
     // a specialized custom gate is of the form: "cz q0 q3"
+    QL_DOUT("is specialized custom gate availablefor for instr " << instr);
```
wvlothuizen: typo

src/ql/ir/compat/platform.cc
```diff
@@ -460,6 +460,7 @@ void Platform::load(
     const utils::Str &platform_config_fname,
     const utils::Str &compiler_config
 ) {
+    QL_DOUT("compat::platform::load");
```
wvlothuizen: "compat::platform" seems a bit redundant since QL_DOUT already prints full file name

src/ql/ir/new_to_old.cc
```diff
@@ -801,18 +801,30 @@ NewToOldConverter::NewToOldConverter(const Ref &ir) : ir(ir) {
...
     } else {
         old_platform = compat::Platform::build(
             ir->platform->name,
             ir->platform->data.data
         );
+        QL_DOUT("NewToOldConverter: got old_platform by building it (compat::Platform::build) from new platform data");
     }
+    QL_DOUT("NewToOldConvertor old instruction_map:");
+    QL_DOUT("(disabled)");
+//    QL_IF_LOG_DEBUG {
```
wvlothuizen: use #if/#else instead of comments to choose between disabled and active. Does this have generic use?

src/ql/ir/old_to_new.cc
```diff
@@ -921,7 +921,8 @@ Ref convert_old_to_new(const compat::PlatformRef &old) {
 
     // Check the result.
     QL_DOUT("Result of old->new IR platform conversion:");
-    QL_IF_LOG_DEBUG(ir->dump_seq());
+    QL_DOUT("(disabled)");
```
wvlothuizen: use #if/#else. I guess what we really need is finer control over log output per 'module' like Python has

src/ql/ir/old_to_new.cc
```diff
@@ -1687,8 +1691,10 @@ Ref convert_old_to_new(const compat::ProgramRef &old) {
 
     // Check the result.
     QL_DOUT("Result of old->new IR program conversion:");
-    QL_IF_LOG_DEBUG(ir->dump_seq());
+    QL_DOUT("(disabled)");
+    // QL_IF_LOG_DEBUG(ir->dump_seq());
```
wvlothuizen: see above

src/ql/resource/instrument.cc
```diff
@@ -464,33 +464,33 @@ void InstrumentResource::on_initialize(rmgr::Direction direction) {
...
-        std::cout << "1q_instr: " << config->single_qubit_instruments << std::endl;
-        std::cout << "2q_instr_q0: " << config->two_qubit_instrument[0] << std::endl;
-        std::cout << "2q_instr_q1: " << config->two_qubit_instrument[1] << std::endl;
-        std::cout << "2q_instr_edge: " << config->two_qubit_edge_instrument << std::endl;
-        std::cout << "nq_instr_q0: " << config->multi_qubit_instrument[0] << std::endl;
-        std::cout << "nq_instr_q1: " << config->multi_qubit_instrument[1] << std::endl;
-        std::cout << "nq_instr_qn: " << config->multi_qubit_instrument[2] << std::endl;
-        std::cout << "====================================" << std::endl;
-    }
+//    QL_IF_LOG_DEBUG {
```
wvlothuizen: disable using #if, not comments

src/ql/resource/inter_core_channel.cc
```diff
@@ -185,26 +185,26 @@ void InterCoreChannelResource::on_initialize(rmgr::Direction direction) {
...
-        std::cout << "preds[1q]: " << config->predicates[0] << std::endl;
-        std::cout << "preds[2q]: " << config->predicates[1] << std::endl;
-        std::cout << "preds[nq]: " << config->predicates[2] << std::endl;
-        std::cout << "inter_core_required: " << config->inter_core_required << std::endl;
-        std::cout << "communication_qubit_only: " << config->communication_qubit_only << std::endl;
-        std::cout << "num_cores: " << config->num_cores << std::endl;
-        std::cout << "num_channels: " << config->num_channels << std::endl;
-        std::cout << "====================================" << std::endl;
-    }
+//    QL_IF_LOG_DEBUG {
```
wvlothuizen: see above

jvansomeren: Updated the pattern of QL_IF_LOG_DEBUG use and definition.

src/ql/ir/old_to_new.cc
jvansomeren: Yes, agreed.
Don't know solution yet for this.
Updated the idiom to simulate enable/disable per file.
This allows finer control over endless dumping, at least
because the debug output got clogged up by those dumps.

src/ql/com/topology.cc
```diff
@@ -7,6 +7,10 @@
 
 #include "ql/utils/logger.h"
 
+// comment two lines below out to enable LOG_DEBUG ir dumping
```
wvlothuizen: Redefining a 'system' macro isn't good practice in my opinion. For one thing, the name no longer reflects the functionality, and although the compiler will probably remove unreachable code, it provides a runtime solution for a compile time choice.

I think it's cleaner to use:
```
#define DEBUG_VERBOSE 1

#if DEBUG_VERBOSE
...
#else
...
#endif
```

src/ql/pass/sch/list_schedule/list_schedule.cc
```diff
@@ -210,9 +214,11 @@ void ListSchedulePass::run_on_block(
     }
 
     // Always dump dot for the schedule if we're debugging.
-    if (QL_IS_LOG_DEBUG) {
-        QL_COUT("dumping dot file...");
+    QL_IF_LOG_DEBUG {
```
wvlothuizen, line 217: This makes the decision compile time instead of runtime

src/ql/resource/instrument.cc
```diff
@@ -5,10 +5,11 @@
 
 #include "ql/resource/instrument.h"
 
+// comment lines below out to enable LOG_DEBUG ir dumping
 /*#undef QL_DOUT
```
wvlothuizen: remove orignal hack

src/ql/com/topology.cc
jvansomeren: Yes, you're right.
Still the current solution (with #ifdef, see the commit message) doesn't seem right either.
I would prefer (a family of) options to turn this on/off per file (not per class or whatever because that is not obvious to the debugger), and to revise the QL_*OUT with that.
In CoSy we had a standard 'engine' (is pass here) option called 'debug' which has a level as value (similar to log_level) which requires QL_DOUT to be sensitive to a pass-level variable set by this option. The latter requires C++ expertise to implement here. The pass level is not appropriate in OpenQL because there is so much common code (not in pass) although we could interpret 'pass' dynamically, i.e. QL_DOUT encountered between the start and stop of a particular pass, which would include the common code too when it is executed on behalf of that pass.
This should be a separate design and not pursued in this current PR.

jvansomeren:
Yes, you're right with your comments, hence a new idiom ...

Still the current solution (with #ifdef, see the commit message) doesn't seem right either.
I would prefer (a family of) options to turn this on/off per file (not per class or whatever because that is not obvious to the debugger), and to revise the QL_*OUT with that.
In CoSy we had a standard 'engine' (is pass here) option called 'debug' which has a level as value (similar to log_level) which requires QL_DOUT to be sensitive to a pass-level variable set by this option. The latter requires C++ expertise to implement here. The pass level is not appropriate in OpenQL because there is so much common code (not in pass) although we could interpret 'pass' dynamically, i.e. QL_DOUT encountered between the start and stop of a particular pass, which would include the common code too when it is executed on behalf of that pass.
This should be a separate design and not pursued in this current PR.

src/ql/com/topology.cc
```diff
@@ -7,6 +7,9 @@
 
 #include "ql/utils/logger.h"
 
+// #define MULTI_LINE_LOG_DEBUG to enable multi-line dumping 
+#undef MULTI_LINE_LOG_DEBUG
```
wvlothuizen, line 11: No need to undef, since there's an assumption this has file scope. I'll file an issue t extend the logging system with finer control at run time

wvlothuizen (approved):

src/ql/com/topology.cc
jvansomeren: Ok, I'll update this, commit and merge.

PR #439, opened 2022-03-24
rOEE mapping algorithm implementation.

Hopery:


jvansomeren (changes requested):
Hi Anabel,
Would you please:

1. merge in develop again (I merged in the updated channel resource just now)
2. add an option to turn your code on/off
3. turn the example .cc into a .py (you can use tests/test_multi_core.py as inspiration)

With that, I can compare your implementation against the current one,
whether both pass all tests, etc.
By the way,
python3 -m pytest
executes all tests and they must pass, of course.
Thanks,
Best,
Hans

wvlothuizen:
I'd also like to ask to use a branch with some meaningful name instead of develop, to ease later understanding of git log messages

Hopery:
Close pull request to change branch's name.

PR #443, opened 2022-04-01
Swap decomposition

jvansomeren:
This branch was set up to repair a bug which turned out not to be in the OpenQL code. The bug was in a private config file.
In the course of debugging, several improvements were done to the debugging support (QL_DOUTs, multi-line debugging).
And a test was updated to compare with a golden output.

But we might conclude that OpenQL doesn't test adequately on the contents of strings such as "cz q0 q3" in the config file.
Note that this string is wrong, the comma between the operands is missing.


wvlothuizen:
> But we might conclude that OpenQL doesn't test adequately on the contents of strings such as "cz q0 q3" in the config file.
> Note that this string is wrong, the comma between the operands is missing.

It would be great to also add a solution for that, or file an issue, preferentially with any diagnosis performed on where things go wrong.

jvansomeren:
Yes, that would be great.
Solving it must be done in a separate PR.

But to add a test that shows it now, requires a modified config file.
I can imagine adding a test in the repository: tests/wrong_config_file.py that reads tests/wrong_config_file.json that will certainly fail,
but I don’t like to pollute the repository with this.

After solving the issue, we probably can make a test (or set of tests) that each check a particular problem in the config file and fail on that.
That failing is then OK, not failing is not OK.
This requires some python scripting.

Analyzing the full extend of the problem to make it an issue, is also some work.
Things that I doubt now, are (all in the old config file):

In the decomposition section:

  *   key: “swap q0 q1” (specialized decomposition)
  *   key: “swap %0 %1” (parameterized decomposition); key: “swap” (parameterized decomposition) is ok
  *   subinstruction in the decomposition: “cz q0 q1”
  *   subinstruction in the decomposition: “cz %0 %1”

In the instructions section:

  *   key: “cz q0 q1” (specialized instruction)
  *   key: “cz %0 %1” (parametrized instruction)

That these are not diagnosed when present, I still need to verify.
I was testing with a config file where all of these were present.
Then replaced all the spaces between operands by “,”, and this helped.

Any suggestions?

At least I can file an issue based on the analysis and text above.

Op 1 apr. 2022, om 16:56 heeft Wouter Vlothuizen ***@***.******@***.***>> het volgende geschreven:



But we might conclude that OpenQL doesn't test adequately on the contents of strings such as "cz q0 q3" in the config file.
Note that this string is wrong, the comma between the operands is missing.

It would be great to also add a solution for that, or file an issue, preferentially with any diagnosis performed on where things go wrong.

wvlothuizen (approved):
Approving this PR, further work happens in the newly created issue #444

Issue #448, opened 2022-04-14
Support for parameterized instructions


wvlothuizen:
In https://github.com/QuTech-Delft/OpenQL/issues/419, support for parameterized gate **decompositions** was tracked. It was concluded that such support already exists and works, but is only meaningful if the parameter can somehow be consumed, either by supporting parameterized gates at the backend level, or by allowing the following type of decompositions to work (with a much longer if-chain in practice):
```
// Parameterized gate decompositions (test)
        "_rx": {
           "prototype": ["X:qubit", "L:int"],
           "duration": @MW_DURATION@,
           "decomposition": {
               "into": [  // FIXME: makes no real sense, and currently fails in backend
                    "if (op(1) < 45) {",
                    "   rx45 op(0)",
                    "} else {",
                    "   rx90 op(0)",
                    "}"
                ]
           }
```

The latter approach currently fails because the backend doesn't support constant expressions (e.g. `23 < 45`), and that is the case because it was earlier decided that such support would be better offloaded to a separate pass, instead of polluting the backend (see https://github.com/QuTech-Delft/OpenQL/issues/424). Such a pass wouldn't only support paramerized instructions, but also adds generic support for constant expressions in other contexts.

wvlothuizen:
Implemented by PR #449

PR #449, opened 2022-04-26
Constant propagation & dead code removal

wvlothuizen:
To cater for #448, this PR provides two new passes: constant propagation and dead code removal. In addition to support for parameterized instructions, this indirectly also adds support for constant expressions to the CC.
Since these are separate passes, they may also be useful in other contexts than the CC backend.


jvansomeren:
Would having a visitor of references to IR nodes made this simpler?
Replacement is not just reading.
Groet,

Hans van Someren



Op 18 mei 2022 om 11:36 heeft Wouter Vlothuizen ***@***.***> het volgende geschreven:

﻿

@wvlothuizen<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_wvlothuizen&d=DwMCaQ&c=XYzUhXBD2cD-CornpT4QE19xOJBbRy-TBPLK0X9U2o8&r=kNdT9ewT6pQdYFkBLR_5-ZqsrSTk7k5Hdd7MSC_Vnzg&m=tegB_vzC9PkeTkqbNp_B-PA3SX1skTF73-ED50cqEeROwi4BLEJ6w-LtYOcmq9sA&s=cIXFrsywsNv6jtdJRy8-aKDOhU8ytQuYTF7WZo5ArVY&e=> requested your review on: #449<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_QuTech-2DDelft_OpenQL_pull_449&d=DwMCaQ&c=XYzUhXBD2cD-CornpT4QE19xOJBbRy-TBPLK0X9U2o8&r=kNdT9ewT6pQdYFkBLR_5-ZqsrSTk7k5Hdd7MSC_Vnzg&m=tegB_vzC9PkeTkqbNp_B-PA3SX1skTF73-ED50cqEeROwi4BLEJ6w-LtYOcmq9sA&s=xvAXWPy3IfTh_qZG-1NZqNmU2llfmnEME_YwDr37HQE&e=> Constant propagation & dead code removal.

wvlothuizen:
@jvansomeren what does "this" refer to, can you elaborate (using the Github web page if possible, mail replies look a bit messy)?

src/ql/pass/opt/const_prop/detail/propagate.cc
```diff
@@ -0,0 +1,224 @@
...
+                    key += "i";
+                } else if (operand->as_bit_literal()) {
+                    key += "b";
+                } else {
+                    DEBUG("not touching operand '" << ir::describe(operand) << "'");
+                    return;     // we don't handle functions with other operand types, but leave them untouched
+                }
+            }
+
+            // NB: we don't perform type promotions like libqasm's cQASM resolver, see FunctionTable::call
```
jvansomeren, line 167: Why not?
I saw the int function. So why not 'real', 'bool', etc. as well?
Or, what do you mean by 'promotion'? Automatic conversions inserted by libqasm, coercions as Jeroen calls them? I didn't investigate but I assume that libqasm makes these explicit and adds them to the AST.

jvansomeren, line 167: Why not?

tests/cc/test_cc_cqasm.py
```diff
@@ -31,42 +31,45 @@ def run_test_case(self, name):
...
-                        'debug': 'yes'
-                    }
-                )
-
             if 1:
                 # use pass manager
                 pl = ql.Platform("cc", "config_cc_s17_direct_iq_openql_0_10.json")
                 c = pl.get_compiler()
 
+                # insert passes at front (in reverse run order)
```
jvansomeren, line 39: Could you make deadcodeelim and propagate passes that need not be configured in this way and are called by default?

jvansomeren, line 39: Could you make deadcodeelim and propagate passes that need not be configured in this way and are called by default?

tests/cc/test_const_prop.cq
```diff
@@ -0,0 +1,102 @@
...
+
+map i = creg(0)
+map j = creg(1)
+
+.condConstExpr
+while (4==4) {
+}
+
+.deadCodeElim
+if (3<2) {
```
jvansomeren, line 28: Why aren't these constant expressions already folded by libqasm?

jvansomeren, line 28: Why aren't these constant expressions already folded by libqasm?

tests/cc/test_const_prop.cq
```diff
@@ -0,0 +1,102 @@
...
+#set i = 4294967295  # out of range for (signed) int
+set i = 2147483647  # max signed int
+#set i = 2147483647 + 1 # out of range
+
+.more
+#set i = 4<5
+set i = int(1+4)
+set i = 1+5
+set i = 1+2-3
+set i = int(j+4)
```
jvansomeren, line 102: Here I see conversions. Earlier I read that conversions weren't supported in the constant propagator. I assume they are supported by libqasm. So this works because libqasm evaluates them?

jvansomeren, line 103: Here I see conversions. Earlier I read that conversions weren't supported in the constant propagator. I assume they are supported by libqasm. So this works because libqasm evaluates them?

tests/cc/test_looping.cq
```diff
@@ -189,10 +189,6 @@ while (b[0]) {
 #while (1 < i+3) { # Error: Input error: in pass VQ1Asm, phase main: in block 'condComplexExpr': in expression '1 < creg[0] + 3': cannot handle function call within function call 'creg[0] + 3'
 #}
 
-#.condConstExpr
-#while (4==4) {
-#}
-
 # single block, with structured control flow mid way
 # 'while' loop does not seem to be a barrier for scheduling by itself
 .multiKernel
```
jvansomeren, line 194: Jeroen implemented dependence analysis such that there is not automatically a barrier before and after any loop. I haven't reviewed that well enough and don't understand how that could be possible. I would have expected that before and after each loop body (so inside the while) there would be a barrier. This would cater for taking the iteration, i.e. the dependence graph cannot be circular so cannot represent the dependences between two successive iterations.
So, this is not a problem here, but a reminder to investigate this further and to come to this later when that investigation has been done.

jvansomeren, line 194: Jeroen implemented dependence analysis such that there is not automatically a barrier before and after any loop. I haven't reviewed that well enough and don't understand how that could be possible. I would have expected that before and after each loop body (so inside the while) there would be a barrier. This would cater for taking the iteration, i.e. the dependence graph cannot be circular so cannot represent the dependences between two successive iterations.
So, this is not a problem here, but a reminder to investigate this further and to come to this later when that investigation has been done.

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
```diff
@@ -0,0 +1,174 @@
...
+                        block->statements.add(st, stmt_idx++);  // NB: increment affects both this add and statement loop
+                    }
+                }
+
+                remove_if_else = true;
+            }
+
+            // if needed, remove the complete if_else statement altogether
+            if (remove_if_else) {
+                block->statements.remove(if_else_idx-1);    // FIXME: why is -1 necessary?
```
jvansomeren: Is the reason a bug in remove or symmetry with add (add adds after the index, so remove would remove after the index). In linked lists you want to know the node before to remove the link to the current one.
So, have you found the reason?

jvansomeren: Is the reason a bug in remove or symmetry with add (add adds after the index, so remove would remove after the index). In linked lists you want to know the node before to remove the link to the current one.
So, have you found the reason?

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
```diff
@@ -0,0 +1,174 @@
...
+            }
+
+        // handle loop
+        } else if (auto loop = statement->as_loop()) {
+            // descend loop body
+            run_on_block(loop->body);
+
+            // NB: we currently have no real use for optimizing static loops, note that we cannot fully
```
jvansomeren, line 140: A loop doing only one iteration can be optimized out: can be replaced by the body.
A break would jump to the end of the body. A continue too. The iteration variable deserves special attention because in a single iteration loop, the break would leave the variable unchanged and the continue would still execute the increment.

jvansomeren, line 140: A loop doing only one iteration can be optimized out: can be replaced by the body.

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
```diff
@@ -0,0 +1,174 @@
...
+            // remove loop anyway if break or continue exists
+        }
```
jvansomeren, line 142: It would be nice to convert a C-style for loop to a foreach loop, when possible. Of a foreach loop it is easier to compute the iteration count, and that is something that passes may like to do.
But this only becomes relevant when OpenQL would unroll loops and other loop optimizations, and that is far away.

jvansomeren, line 142: It would be nice to convert a C-style for loop to a foreach loop, when possible. Of a foreach loop it is easier to compute the iteration count, and that is something that passes may like to do.
But this only becomes relevant when OpenQL would unroll loops and other loop optimizations, and that is far away.

src/ql/pass/opt/const_prop/detail/propagate.cc
```diff
@@ -0,0 +1,224 @@
...
+private:    // functions
+
+    /*
+     * Handle an expression node, i.e. replace eligible functions calls with a literal expression.
+     *
+     * Note that we cannot directly use visit_expression(), because that has a 'ir::Expression &' as parameter.
+     * It is therefore not possible to change a function_call into a (say) int_literal. Here the expression is
+     * wrapped in a 'utils::One', allowing polymorphic access. This does force us however to visit all relevant node
+     * types that contain an expression (although we can skip e.g. custom_instruction.operands).
+     */
```
jvansomeren, line 145: I'm still puzzling on this. This need requires a particular convention (all those '*'s) which is annoying. Another kind of visitor that visits the nodes that contain expressions would avoid this; it would also make deleting nodes easier.
And also the cqasm reader and cqasm writer are not written with a visitor, illustrating to me that the current visitor is not powerful enough. So I'd expect a family of visitors of which the current visitor is the simplest one, the visitor we actually require here is the one-but-simplest, and the cqasm reader one of the hardest. In CoSy we also had context visitors (where the visit stack was accessible as a parameter), which allowed to visit depending on attributes of nodes higher up in the IR-tree (e.g. top if-then-else, or if-then-else inside another one or inside a loop).
Have you considered changing the visit function instead of applying the current solution?

jvansomeren, line 145: I'm still puzzling on this. This need requires a particular convention (all those '*'s) which is annoying. Another kind of visitor that visits the nodes that contain expressions would avoid this; it would also make deleting nodes easier.
And also the cqasm reader and cqasm writer are not written with a visitor, illustrating to me that the current visitor is not powerful enough. So I'd expect a family of visitors of which the current visitor is the simplest one, the visitor we actually require here is the one-but-simplest, and the cqasm reader one of the hardest. In CoSy we also had context visitors (where the visit stack was accessible as a parameter), which allowed to visit depending on attributes of nodes higher up in the IR-tree (e.g. top if-then-else, or if-then-else inside another one or inside a loop).
Have you considered changing the visit function instead of applying the current solution?

src/ql/pass/opt/const_prop/detail/platform_functions.h
```diff
@@ -0,0 +1,54 @@
...
+ * - P_B: bit literal
+ * - P_I: integer literal
+ *
+ * Based on:
+ * - deps/libqasm/src/cqasm/func-gen/funcgen.cpp
+ * - <build>/deps/libqasm/src/cqasm/cqasm-v1-functions-gen.cpp
+ * - src/ql/arch/cc/pass/gen/vq1asm/detail/functions.cc
+ *
+ * Note: to have better guarantees on consistency, we might also use this file in ir/old_to_new.cc to create the
+ * platform functions, and extend it to be usable for ir/operator_info.cc
```
jvansomeren, line 24: Somehow this set of functions is not a proper part of the IR now.
More and more passes will need this list so standardization is required.
And it needs to be at least a superset of what all OpenQL targets support.
And then there needs to be pass for each target that checks and enforces the entry conditions of the code generator (which can have run-time equivalents, etc.).
There maybe even functions to which calls are created by the compiler and that are not created because calls were found in a cqasm file; this certainly is true when you consider that there are multiple cqasm versions, but also when another language (e.g. OpenQASM) would be read by OpenQL.
So the function list needs to be standardized as part of the IR of OpenQL; and we need to be able to distinguish such standard functions from regular functions.
I guess the above is not comment on this code but deserves a proper issue.

jvansomeren, line 24: Somehow this set of functions is not a proper part of the IR now.
More and more passes will need this list so standardization is required.
And it needs to be at least a superset of what all OpenQL targets support.
And then there needs to be pass for each target that checks and enforces the entry conditions of the code generator (which can have run-time equivalents, etc.).
There maybe even functions to which calls are created by the compiler and that are not created because calls were found in a cqasm file; this certainly is true when you consider that there are multiple cqasm versions, but also when another language (e.g. OpenQASM) would be read by OpenQL.
So the function list needs to be standardized as part of the IR of OpenQL; and we need to be able to distinguish such standard functions from regular functions.
I guess the above is not comment on this code but deserves a proper issue.

src/ql/pass/opt/const_prop/detail/propagate.cc
wvlothuizen: The basic reason is that it currently makes little sense since OpenQL only supports a subset of the types libqasm knows about. We could promote a bool to an int when possible, but since (at least in the CC) int's and bool's live in different memory spaces requiring an explicit int cast by the user seems a better solution

tests/cc/test_cc_cqasm.py
wvlothuizen: I could, but do note that the default pass list is set globally, and was historically intended to give behaviour that's compatible with older versions. And in practice manual control is needed anyway in most cases.

wvlothuizen: see above

tests/cc/test_const_prop.cq
wvlothuizen: Quoting from `arch::cc::pass::gen::vq1asm::detail::codegen.h`, especially see that last sentence:

   /**
     * Handle expressions.
     *
     * To understand how cQASM functions end up in the IR, please note that functions are handled during analyzing
     * cQASM, see 'AnalyzerHelper::analyze_function()'.
     *
     * A default set of functions that only handle constant arguments is provided by libqasm, see
     * 'register_into(resolver::FunctionTable &table)'. These functions add a constant node to the IR when called (and
     * fail if the arguments are not constant)
     *
     * Some of these are overridden by OpenQL to allow use of non-constant arguments. This is a 2 step process, where
     * 'convert_old_to_new(const compat::PlatformRef &old)' adds functions to ir->platform using 'add_function_type',
     * and 'ql::ir::cqasm:read()' then walks 'ir->platform->functions' and adds the functions using
     * 'register_function()'. These functions add a 'cqv::Function' node to the IR (even if the arguments are constant,
     * so overriding a function defeats libqasm's constant removal for that function).
     */

wvlothuizen: see above

tests/cc/test_const_prop.cq
wvlothuizen: also see above

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
wvlothuizen: Not yet, will have a closer look

tests/cc/test_looping.cq
wvlothuizen: duplicate

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
wvlothuizen: duplicate

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
wvlothuizen: Fully agree, but I don't really see use cases where this optimization would currently provide a benefit

wvlothuizen: duplicate

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
wvlothuizen: Agree

src/ql/pass/opt/const_prop/detail/propagate.cc
wvlothuizen: There are multiple ways to traverse the tree depending on needs, see https://tree-gen.readthedocs.io/en/latest/tree-gen.gen.html#tree-traversal, so generally speaking it's not necessarily a bad thing that the cQASM reader/writer don't use a visitor.

I'm not sure whether the need we have here could easily be generalized: one could envision a visitor that doesn't dereference One/Any/etc, but in other places than Expressions that may not be what you want.

wvlothuizen: duplicate

src/ql/pass/opt/const_prop/detail/platform_functions.h
wvlothuizen: Agreed. Functions need a proper redesign anyway when the ideas in https://github.com/QuTech-Delft/cQASM-spec/tree/95ee081c0910548b8201c510f88ebc89be9284e8 are implemented. This would also improve the user experience a lot, because currently gate decomposition works a lot like a function call, but otherwise user defined functions don't exist, so there is no way to reuse pieces of code

wvlothuizen: duplicate

src/ql/pass/opt/const_prop/detail/propagate.cc
jvansomeren: I would then also expect the support of an explicit bit/bool cast. Such a cast then involves moving a value from one to another space and require allocation in that second space.
Do you have a write-up of the programming model explaining the limitations of classical programming for the CC?
Because when the bits live in a space that doesn't allow allocation, the cast wouldn't indeed make sense.

tests/cc/test_const_prop.cq
jvansomeren: Created an issue https://github.com/QuTech-Delft/OpenQL/issues/451 for this

src/ql/pass/opt/const_prop/detail/propagate.cc
wvlothuizen: There is a cast from bit to int (`int(bit b)`) in the standard platform functions, which involves moving between the two memory spaces on the CC. No other casts exist by default, although `operator>` could be used to get a C-style int to bool conversion.
On the CC, bits live in Distributed Shared Memory, which can only be written by memory, not by instructions (or in the STATE register, which **can** be witten to, but which isn't supported yet by the backend)

src/ql/pass/opt/dead_code_elim/dead_code_elim.cc
wvlothuizen: solved! Was an (incorrect) solution for `.add(pos)` adding before pos, not after

jvansomeren (approved):
No more comment

jvansomeren:
Or even a visitor that maintains a stack of context IR nodes? This is overkill for this constant folding but comes in handy when e.g. only array accesses need to be replaced.
I’d like to step back to assess whether providing more intelligent visitors would be helpful. And this is the first one that needs more than just looking into the IR.
Groet,

Hans van Someren



Op 18 mei 2022 om 11:36 heeft Wouter Vlothuizen ***@***.***> het volgende geschreven:

﻿

@wvlothuizen<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_wvlothuizen&d=DwMCaQ&c=XYzUhXBD2cD-CornpT4QE19xOJBbRy-TBPLK0X9U2o8&r=kNdT9ewT6pQdYFkBLR_5-ZqsrSTk7k5Hdd7MSC_Vnzg&m=tegB_vzC9PkeTkqbNp_B-PA3SX1skTF73-ED50cqEeROwi4BLEJ6w-LtYOcmq9sA&s=cIXFrsywsNv6jtdJRy8-aKDOhU8ytQuYTF7WZo5ArVY&e=> requested your review on: #449<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_QuTech-2DDelft_OpenQL_pull_449&d=DwMCaQ&c=XYzUhXBD2cD-CornpT4QE19xOJBbRy-TBPLK0X9U2o8&r=kNdT9ewT6pQdYFkBLR_5-ZqsrSTk7k5Hdd7MSC_Vnzg&m=tegB_vzC9PkeTkqbNp_B-PA3SX1skTF73-ED50cqEeROwi4BLEJ6w-LtYOcmq9sA&s=xvAXWPy3IfTh_qZG-1NZqNmU2llfmnEME_YwDr37HQE&e=> Constant propagation & dead code removal.

PR #454, opened 2022-06-17
Automatic state preparation

anneriet:
This adds a `state_prep([statevector], [qubits])` function to the `Kernel`, which generates and adds gates to the circuit that take the qubits from a |0> state to the state specified in the statevector. It uses some of the functions from unitary decomposition to do this.

I still need to write documentation, and I have some example/test scripts I would love to add to the project but not sure about the best approach there, since most of the python scripts in the /example folder don't currently work?

Lmk what you think, figured a pull request was as good a place to open the conversation as any :P


jvansomeren:
You wrote:

> I still need to write documentation, and I have some example/test scripts I would love to add to the project but not sure about the best approach there, since most of the python scripts in the /example folder don't currently work?

As far as I know, Jeroen did the OpenQL makeover after your last contribution. So please check out the (also new) documentation and especially how to set up the system since that also has changed.
Compiling in unitary decomposition is an option defined in CMakelists.txt but I checked that its default is ON.
To make your tests work, please checkout the most simple test scripts in the /tests directory and update yours accordingly; perhaps your idea of which options are on/off also has changed, as well as the configuration taken.

wvlothuizen:
> Compiling in unitary decomposition is an option defined in CMakelists.txt but I checked that its default is ON.

But it looks like compilation will fail if that option is turned off, since Kernel now calls Unitary. 

Is Kernel the proper place for this functionality anyway, it seems a bit high-level with respect to other Kernel functions. And we now have two input methods: the API and cQasm, adding functionality to Kernel is not really future proof. Maybe a decomposition pass would be an option?

src/ql/api/kernel.i
```diff
@@ -240,6 +240,19 @@ Returns
...
 """
 
+%feature("doctring") ql::api::Kernel::state_prep
+"""
+Automatic state preparation, currently requires unitary decomposition for all cases.
+
+Parameters
+----------
+states : Complex Vector
+    Specifies the state (change) of the qubits after the state preparation
```
pablolh, line 250: Maybe asking a stupid question, but just to be clear: the state will be the same for all the "prepared" qubits right?

src/ql/ir/compat/kernel.cc
```diff
@@ -1055,6 +1055,22 @@ void Kernel::gate(
     gates.extend(u.get_decomposition(qubits));
 }
 
+// adding state prepration / (arbitrary) qubit initialisation to kernel
+void Kernel::state_prep(
+    const Vec<Complex> &array,
+    const Vec<UInt> &qubits
+) {
+    QL_DOUT("Preparing state with array " << array);
+    cycles_valid = false;
```
pablolh, line 1064: Could it be worth it to assert some preconditions on `array`?
e.g. it should be a proper quantum state with sum of square of modules == 1

src/ql/ir/compat/kernel.cc
```diff
@@ -1055,6 +1055,22 @@ void Kernel::gate(
     gates.extend(u.get_decomposition(qubits));
 }
 
+// adding state prepration / (arbitrary) qubit initialisation to kernel
+void Kernel::state_prep(
```
pablolh, line 1059: It's something that is missing in the whole codebase, so not your fault, but this _should_ have an accompanying unit test. I have a PR in progress that adds the GoogleTest framework and some unit tests for passes. I intend to write those for kernel and other components as well.

src/ql/api/unitary.i
```diff
@@ -28,7 +28,6 @@ Returns
 None
 """
 
-
```
pablolh, line 31: Not needed

pablolh, line 31: Not needed

src/ql/com/dec/unitary.cc
```diff
@@ -665,6 +666,14 @@ static Int recursiveRelationsForUnitaryDecomposition(
     }
 }
 
+/**
+ * Does state preparation, results in a circuit for which A|0> = |psi> for given state psi, stored as the array in Unitary
+ */
+ir::compat::GateRefs Unitary::prepare_state(const utils::Vec<utils::UInt> &qubits){
+    QL_EOUT("State preparation not implemented!");
```
pablolh: Since you do a PR with multiple commits, you might as well squash all the commits together and clean up so that the history of the main branch is clean after the merge.

src/ql/api/kernel.i
anneriet: The qubits will be in a state as specified by the vector, which (depending on the vector) can be something different for each of the qubits. (The state vector for |01> is [0 1 0 0]^T for example, so the first qubit is state |0> and the second is in state |1>) The state preparation takes the vector and uses it to prepare the total combined state on the specified qubits.

src/ql/ir/compat/kernel.cc
anneriet: I was planning to make it do auto-normalization, because doing that manually becomes annoying quickly, but I dont remember if I actually implemented it. If I didnt, then there should be a check for it indeed!

src/ql/ir/compat/kernel.cc
anneriet: Fair and completely correct, I wish you success! And if you need me to write some tests for this code, or the general unitary decomposition, let me know, Id be happy to help.

src/ql/com/dec/unitary.cc
```diff
@@ -355,24 +331,25 @@ class UnitaryDecomposer {
...
 
-        Complex det = matrix.determinant();// matrix(0,0)*matrix(1,1)-matrix(1,0)*matrix(0,1);
+    //     Complex det = matrix.determinant();// matrix(0,0)*matrix(1,1)-matrix(1,0)*matrix(0,1);
+    Vec<Real> zyz_decomp(Complex A, Complex B, Complex det){
+        Vec<Real> zyz_angles; // gamma, beta, alpha
 
-        Real delta = atan2(det.imag(), det.real())/matrix.rows();
-        Complex A = exp(Complex(0,-1)*delta)*matrix(0,0);
-        Complex B = exp(Complex(0,-1)*delta)*matrix(0,1); //to comply with the other y-gate definition
+        Real delta = 0.5*atan2(det.imag(), det.real());
```
pablolh, line 343: The absence of a single unit test for all this math gets me very worried

src/ql/com/dec/unitary.cc
```diff
@@ -665,6 +624,88 @@ static Int recursiveRelationsForUnitaryDecomposition(
     }
 }
 
+/**
+ * Does state preparation, results in a circuit for which A|0> = |psi> for given state psi, stored as the array in Unitary
+ */
+ir::compat::GateRefs Unitary::prepare_state(const utils::Vec<utils::UInt> &qubits){
+    UInt nqubits = qubits.size();
+    if (((UInt) 1) << nqubits != array.size()){
+        QL_FATAL("Length of state prepatation vector does not match number of qubits! Expected vector of size " << (1 << nqubits) << " but got vector of size " << array.size());
```
pablolh, line 648: prepatation -> preparation (typo)

src/ql/com/dec/unitary.cc
pablolh, line 343: This class already has a member called `delta`. You're defining a local variable with the same name as a class member. It seems to me that this class uses the members for computation, but those don't need to be kept between methods. Looked quickly, I think the member delta, alpha, ... could just be local variables.

src/ql/com/dec/unitary.cc
```diff
@@ -665,6 +624,88 @@ static Int recursiveRelationsForUnitaryDecomposition(
...
+            QL_DOUT("Sending indices " << start_i << " until " << (start_i + ngates) << " to multicontrolled z and y. i=" << i);
+            Eigen::CompleteOrthogonalDecomposition<Eigen::MatrixXd> dec(decomposer.genMk_lookuptable[i - 1]);
+            temp = Eigen::Map<Eigen::VectorXd>(theta.data() + start_i, ngates);
+            tr = dec.solve(temp);
+
+            // The first one is always controlled from the next qubit to the last qubit
+            c.emplace<ir::compat::gate_types::RY>(qubits[nqubits-i-1], tr[0]);
+            c.emplace<ir::compat::gate_types::CNot>(qubits[nqubits-i], qubits[nqubits-i-1]);
+            for (UInt j = 1; j < (ngates - 1); j++){
+                idx = i - log2(((j)^((j)>>1))^((j+1)^((j+1)>>1)));
```
pablolh, line 695: It would be nice to extract this kind of computation as an auxiliary function.
- This gives an explicit name to the piece of code, which gives information to the reader
- It makes the code clearer (generally speaking, any function longer than 20 lines should be avoided, IMO)
- It makes it possible to write unit tests for such pieces of code.

More on unit tests:
- They validate the correctness of small pieces of code.
- They allow to check for corner cases easily.
- They provide documentation to the reader, as they showcase what the functions do.
- They provide guarantee that one is not changing the behavior of the code when refactoring.

Test coverage is important to write correct software.

pablolh (changes requested):
This is my first code review with Qutech!

It's difficult for me to review the maths because I'm not so up to date yet.
I'm happy to meet to discuss so you can explain your change.
There is a crucial lack of unit tests in the whole codebase, that doesn't make it easy to understand what the code does and check for its correctness. I will work on fixing that asap.
Especially with this kind of scientific, maths-heavy code, I think it's important to have clean, modular code (small functions doing targetted things, that the compiler would inline, so no performance overhead) with accompanying tests.

It's been mentioned that this PR doesn't have the latest changes in the main branch. So it'd need to be rebased / updated before merging, so we can test if the CI passes.

src/ql/ir/compat/kernel.cc
RichardVersluisTNO: If there isn't already a separate function for (re)normalization in OpenQL, normalization of a unitary matrix should be a separate function (with its own unit test), because it can be useful in many places. Note however that 1) in some situations you want to renormalize "automatically" and in 2) other situations you may want to check if a unitary is (close to) normal, because a non-unitary matrix could indicate an error.

An example of the first situation are rounding errors due to the fixed precision that is used. After some matrix operations you may want to (or have to) renormalize when the Modulus of the matrix drifted beyond a certain delta. 

An example of the 2nd is when the uses supplies a matrix that is expected to be unitary, but you want to check first if it is really (close to) unitary and give an error (too far) or warning (re-normalizing) when it is off.

RichardVersluisTNO: p.s. I once understood that QX emulator does automatic re-normalization of state vectors. This was introduced after it was observed that repeated Hadamard gates did not bring the state vector back to its original state because of the limited precision of sqrt(2) in the operation. Nader then introduced renormalization. I do not were it is in the QX code base but could be used an an inspiration perhaps?

src/ql/api/kernel.i
pablolh: Ah OK it's clearer now, thanks.

src/ql/ir/compat/kernel.cc
pablolh: Very interesting, thanks for the context

pablolh:
> adding functionality to Kernel is not really future proof

Indeed, I think both ways to give input to OpenQL should have similar functionality.
And I don't think "automatic state preparation" will ever be part of cQasm, as it's too high level.

Although this is a really cool feature.
I'm thinking, can't this automatic state preparation be a SEPARATE binary that produces the gates needed to achieve the input state?
As far as I understand, the automatic state prep doesn't do anything that is architecture specific, just adds a bunch of universal gates.
So that instead of running every time you compile, you could once generate some cqasm for state preparation, then paste it into your circuit?

That is just ideas, we can refactor/change later. I have no problem with the code as it is now.
Changing where the state prep is defined/called wouldn't be too much work, I can do it later.

> But it looks like compilation will fail if that option is turned off, since Kernel now calls Unitary

@anneriet did you check that? by adding the `-DWITH_UNITARY_DECOMPOSITION=OFF` option when you call CMake, then check that the code builds?
I actually need to figure out if this is done in CI or not.


@anneriet would also love to see the tests, but we can do a separate merge, as I understand you would rather see this research work in the software now.

pablolh (approved):

PR #455, opened 2022-07-07
[QI2-150] Add support for python 3.10

koffie:
This change makes sure that wheels will also be build for python 3.10


koffie:
The release process fails with this update, so this still needs some work

https://github.com/QuTech-Delft/OpenQL/actions/runs/2629337764

koffie:
Ok, this now works correctly as you can see at:
https://github.com/QuTech-Delft/OpenQL/actions/runs/2629705336

That run still failed in the publish step because apparently the credentials of the pypi user that are stored in `${{ secrets.PYPI_TOKEN }}` are no longer valid. But that has nothing to do with the changes introduced in this PR.

wvlothuizen (approved):
Looks good!

PR #461, opened 2022-09-29
Static self-registration of passes and architectures

pablolh:
Copy of commit message:

    Use static fields in the pass and arch derived classes
    to register them to the factory, instead of having the factories
    register every class.
    
    For this, a CMake OBJECT library has to be used, otherwise the linker
    will get rid of those seemingly unused symbols.
    
    Having the static bool field in the base classes would require the use
    of the Curiously Recurring Template Pattern and a bigger change. I
    decided to leave it for now.


jvansomeren (changes requested):
While going over the files, I only looked at code and forgot to look at comment.
Have you updated the comment that is used for the generation of documentation?
Can you please send me the updated documentation and point out where the new stuff is?
Have you also added tests for this (when possible)?

pablolh:
Indeed I forgot to update the documentation. This is done now.
I haven't added tests; but the existing tests cover the fact that passes are registered (I've experienced that when debugging).

jvansomeren (approved):
As mailed, I'd like to read the new documentation.
Also, with regard to Jeroen's comment, I'd like to be assured that backend compiler users can add completely new passes (with self registration now) and use those in their compiler.

PR #462, opened 2022-10-04
Move visualization passes to new IR

pablolh:
Works in src/ql/pass/ana to remove usage of the old IR.

    - Replace base classe ProgramAnalysis with new-IR base class
    Analysis;
    - New IR already "schedules sequentially" by default (even before
    actual scheduling). See ir/cqasm/read.cc;
    - New IR doesn't contain default gates, therefore remove the default
    visualizations for those;
    - In new IR, durations are expressed in cycles, not in time units.
    This allows some simplifications in the logic;
    - Fix visualizer examples python files using wrong config.
    
    This PR was tested with all possible examples in
    tests/visualizer. Output is identical to what was produced before.


jvansomeren (commented):
Why was deps/cimg/include/CImg.h updated?

pablolh:
The previous deps/cimg/include/CImg.h was old and generated a TON of compiler warnings with my (recent) compiler.
New version doesn't. It's taken from https://github.com/GreycLab/CImg.

I'm wondering if we can instead use a submodule like the other dependencies. The issue I think is that CImg repo (https://github.com/GreycLab/CImg) has some more files that we're not interested in. Wondering if git supports partial submodule download.

src/ql/pass/ana/visualize/detail/circuit.cc
```diff
@@ -756,9 +717,13 @@ CircuitLayout parseCircuitConfiguration(Vec<GateProperties> &gates,
     };
 
     Map<Str, VisualParameters> parameterMapping;
-    for (auto it = platformInstructions.begin(); it != platformInstructions.end(); ++it) {
-        Str gateName = it.key();
-        Json instruction = *it; //.value();
+    for (const auto& instruction: platform.instructions) {
+        Str gateName = instruction->name;
+
+        auto isGateUsed = std::find_if(gates.begin(), gates.end(), [gateName](const GateProperties& g) { return g.name == gateName; }) != gates.end();
```
jvansomeren, line 723: This looks like new functionality. Why was it included?

src/ql/pass/ana/visualize/detail/common.cc
```diff
@@ -18,85 +19,95 @@ namespace detail {
 
 using namespace utils;
 
-Vec<GateProperties> parseGates(const ir::compat::ProgramRef &program) {
-    Vec<GateProperties> gates;
+class GateCollector : public ir::RecursiveVisitor {
```
jvansomeren, line 22: I'd like to assess that a new IR circuit is scheduled always.

src/ql/pass/ana/visualize/detail/mapping.cc
```diff
@@ -275,16 +265,16 @@ void computeMappingPerCycle(const MappingGraphLayout &layout,
...
     auto size = platform->topology->get_grid_size();
     if (size.x == 0 || size.y == 0) {
         QL_IOUT("Falling back on basic visualization; missing qubit coordinates in topology section");
         return false;
     }
     topology.xSize = size.x;
     topology.ySize = size.y;
-    topology.vertices.resize(platform->qubit_count, { 0, 0 });
-    for (utils::UInt q = 0; q < platform->qubit_count; q++) {
+    topology.vertices.resize(platform->qubits->shape[0], { 0, 0 });
```
jvansomeren, line 276: Need to discuss this. Does this work for any number of qubits? I'd expected a scan over the objects of qubits type and collecting them in some vector.

src/ql/pass/map/qubits/map/detail/past.cc
```diff
@@ -602,8 +602,8 @@ void Past::make_real(const ir::compat::GateRef &gate, ir::compat::GateRefs &circ
 
     if (gate->swap_params.part_of_swap) {
         QL_DOUT("original gate was swap/move, adding swap/move parameters for gates in decomposed circuit");
-        for (ir::compat::GateRef &gate : circuit) {
-            gate->swap_params = gate->swap_params;
+        for (ir::compat::GateRef &it : circuit) {
```
jvansomeren, line 605: Strange that the old code was working.

jvansomeren (approved):
No breaking problems.

src/ql/pass/ana/visualize/detail/circuit.cc
pablolh: Indeed it is new functionality.

It prevents unnecessary warning messages that the visualization pass cannot find a visual_type for a gate.

One particular case where this happens is when you use "decomposable" gates. The platform.instructions still contains the "decomposable" gate name, but those gates don't actually occur in the circuit after decomposition. Hence there is no need to define a visual_type for them in the config.

Should you not find a visual_type for a gate that is actually used, you would get the explicit warning
`Warning: Did not find visual for gate: 'x45' with visual_type x45, skipping gate!`
Maybe we should actually error in that case, it's debatable.

src/ql/pass/ana/visualize/detail/common.cc
pablolh: There is a function in src/ql/ir/consistency.cc that is called "check_cycles_non_decreasing" and that checks that all instructions in a new-IR circuit have non-decreasing cycle numbers.

The invariant is also described in ir.tree:
`a program that has not yet been scheduled can be represented with any non-decreasing cycle assignment`

I looked at the reading pass and it does "naively schedule" all instructions in consecutive cycles.
An interesting point is that should this "naive scheduling" actually care about instruction duration? To me this in `read.cc` is not proper:
```
                    // If scheduling information is discarded, increment the
                    // cycle number at the end of each instruction.
                    if (options.schedule_mode != ScheduleMode::KEEP) {
                        cycle++;
                    }
```

as well as this:
```
            // The cycle counter increments at the end of each bundle if
            // scheduling information is retained. Otherwise it is incremented
            // at the end of each instruction.
            if (options.schedule_mode == ScheduleMode::KEEP) {
                if (!ql_bundle.empty() || !found_pragma) {
                    cycle++;
                }
            }
```

I think cycle should be incremented by the duration of the instruction. When there is a bundle, cycle should be incremented by the max duration of the contained instruction. But that's a different change.

To come back to "a new IR circuit is scheduled always", is there no value in displaying a non-scheduled circuit? Considering the invariant.

src/ql/pass/map/qubits/map/detail/past.cc
pablolh: Indeed. I haven't digged into the mapper yet to understand what was happening.

src/ql/pass/ana/visualize/detail/mapping.cc
pablolh: As far as I understand, there was only one object of qubit type, referred to by platform->qubits, and this object should have the dimension of a vector (hence the shape[0]). See:
```
    # The main qubit register that the generic mapper will map everything to
    # and that topology applies to. The data type must be a vector of qubits.
    # This also indirectly defines the main qubit type.
    qubits: Link<physical_object>;
```

OR does the "main qubit register" formulation and your comment imply that there can be multiple?

I checked old_to_new.cc and new_to_old.cc and both seem to pair the number of qubits to platform->qubit_count (for old IR) and the 1st dimension of platform->qubits (for old IR) to the "qubit_number" json value in the platform config file. So in any case, in current code, `platform->qubit_count` should be equal to `platform->qubits->shape[0]`, I believe.

PR #463, opened 2022-10-17
Update visualizer docs after 9447d10d

pablolh:
Visualizer passes now use the new IR; reflect this in the docs.


jvansomeren (approved):
Ok

PR #466, opened 2022-11-09
Make initial placer a separate new-IR pass using new MIP solver

pablolh:
This pretty large commit does a number of inter-dependent things, required to make initial placement great again:

- Separate the initial placer from the existing mapper. This is done by udpating all qubit references in the IR by their corresponding mapped real qubit computed by the MIP solver.
- Use the new IR in this new inital placement pass.
- Rework the code to remove debug logs, which were half-commented and really unpleasant to read.
- Move the MIP solving to use a new solver called HiGHS. This is an award-winning solver that is actively maintained, performant, open source and licensed under the MIT license, which allows the removal of the toggles to enable/disable inital placement due to license issues.
- Remove GLPK dependency.
- Add HiGHS as submodule.
- Add doctest as submodule, and add unit tests for the linear solver. To make unit-testing convenient, the implementation class is independent from the IR and only takes the 2q gates count as input.
- Update documentation.


include/ql/ir/prim.h
```diff
@@ -331,6 +331,7 @@ std::ostream &operator<<(std::ostream &os, const Matrix<T> &mat) {
     return os;
 }
 
+#undef IGNORE
```
pablolh: There is an IGNORE macro defined in a Windows header winbase.h 😮 
Windows 😭

jvansomeren: I would prefer to redefine our IGNORE
Now all users of this header file must check that they don't use a facility that on windows might depend on the setting of this IGNORE macro

src/ql/pass/map/qubits/map/detail/mapper.h
```diff
@@ -117,8 +117,8 @@ std::ostream &operator<<(std::ostream &os, PathStrategy p);
  *    not be mapped to a real qubit yet, and it's possible for real qubits to
  *    be unused. Virtual qubits are in use as soon as they have been encountered
```
jvansomeren: According to what you explained, virtual qubits are not allocated on first use. They start off one-to-one.

src/ql/pass/map/qubits/map/map.cc
```diff
@@ -403,11 +351,8 @@ pmgr::pass_types::NodeType MapQubitsPass::on_construct(
     // Build the options structure for the mapper.
     parsed_options.emplace();
 
-    parsed_options->initialize_one_to_one = options["initialize_one_to_one"].as_bool();
```
jvansomeren, line 406: What happens to old application code (e.g. in tests) after the merge that throws this deleted option to the mapper?

src/ql/pass/map/qubits/place_mip/detail/impl.cc
```diff
@@ -0,0 +1,453 @@
+/** \file
+ * Initial placement engine.
+ */
+
+#include "impl.h"
+
+#include <thread>
```
jvansomeren: Thread still needed?

src/ql/pass/map/qubits/place_mip/detail/impl.cc
```diff
@@ -0,0 +1,453 @@
...
+#include <mutex>
```
jvansomeren: Same for mutex and condition variables?

src/ql/pass/map/qubits/place_mip/detail/impl.cc
```diff
@@ -0,0 +1,453 @@
...
+    model->lp_.sense_ = ObjSense::kMinimize;
+    model->lp_.offset_ = 0;
+
+    model->lp_.col_cost_ = std::vector<double>(model->lp_.num_col_, 0.);
+    std::fill(model->lp_.col_cost_.begin() + nfac * qubitsCount, model->lp_.col_cost_.end(), .1);
+
+    model->lp_.col_lower_ = std::vector<double>(model->lp_.num_col_, 0.);
+
+    model->lp_.col_upper_ = std::vector<double>(model->lp_.num_col_, 1.);
+    std::fill(model->lp_.col_upper_.begin() + nfac * qubitsCount, model->lp_.col_upper_.end(), 1.0e5);
```
jvansomeren: What is 1.0e5?

src/ql/pass/map/qubits/place_mip/detail/impl.cc
```diff
@@ -0,0 +1,453 @@
...
+            for (UInt j = 0; j < nfac; ++j) {
+                for (UInt l = 0; l < qubitsCount; ++l) {
+                    QL_ASSERT(distanceProvider(k, l) < utils::MAX && "All qubits in the topology should be connected");
+                    double newValue = refcount[i][j] * distanceProvider(k, l);
+                    if (j == i && l == k) {
+                        newValue += costmax[i][k];
+                    }
+
+                    // This describes a sparse matrix; 0s don't have to be added.
+                    if (std::abs(newValue) > EPSILON) {
```
jvansomeren, line 222: Is there a relation between EPSILON and 1.0e5?

src/ql/pass/map/qubits/place_mip/detail/impl.cc
```diff
@@ -0,0 +1,453 @@
...
+            }
+        }
+    }
+
+    return costmax;
+}
+
+std::unique_ptr<HighsModel> Impl::createHiGHSModel(const Vec<Vec<UInt>> &refcount) {
+    // This describes the MIP problem using HiGHS's API. Some examples can be found in deps/highs/examples.
+
```
jvansomeren, line 106: What I miss is the MIP problem formulation and a very clear description how it is translated to the API as was in the old implementation.

src/ql/pass/map/qubits/place_mip/detail/impl.h
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Whether to write the MPS model to a file.
+     */
+    utils::Bool write_model_to_file = true;
+
+    /**
+     * Timeout for the MIP algorithm in seconds, or 0 to disable timeout.
+     */
+    utils::Real timeout = 0.0;
```
jvansomeren, line 38: Could you add two additional options that are compatible to the old options that translate to the new options for timeout and horizon?

src/ql/pass/map/qubits/place_mip/detail/impl.h
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * The placement algorithm will only consider the connectivity required to
+     * perform the first horizon two-qubit gate types. 0 means that all
+     * gate types should be considered.
+     */
+    utils::UInt horizon = 0;
+};
+
```
jvansomeren, line 54: Furthermore, the old system had an option to choose what to do on timeout: ignore the mapping and continue, or let the compiler stop. With this new system, it could also be an option to continue with the mapping that is there at the time out.

src/ql/pass/map/qubits/place_mip/place_mip.cc
```diff
@@ -90,81 +93,77 @@ PlaceQubitsPass::PlaceQubitsPass(
...
     const utils::Str &type_name
-) : pmgr::pass_types::KernelTransformation(pass_factory, instance_name, type_name) {
+) : pmgr::pass_types::Transformation(pass_factory, instance_name, type_name) {
     options.add_int(
         "horizon",
         "When specified, the placement algorithm will only consider the "
-        "connectivity required to perform the first N two-qubit gates of each "
-        "kernel. When not specified or zero, all two-qubit gates are "
+        "connectivity required to perform the first N most frequent two-qubit gate types. "
+        "When not specified or zero, all two-qubit gate types are "
```
jvansomeren, line 101: What do you mean by gate type?

src/ql/pass/map/qubits/place_mip/place_mip.cc
```diff
@@ -90,81 +93,77 @@ PlaceQubitsPass::PlaceQubitsPass(
...
         "considered.",
-        "0", 0, 100
+        "0", 0, utils::MAX
+    );
+    options.add_real(
+        "timeout",
+        "A float duration in seconds after which the MIP problem resolution by HiGHS will terminate. "
+        "If this happens, the pass will emit a warning and will not update the qubit indices. "
+        "When set to 0, there is no time limit.",
```
jvansomeren, line 109: So it will not use a mapping that exists at the time out. That is fine. This answers comment above. But there could be an option to stop the compiler. This is useful in benchmarking when silent continuation after a time out would not be desired. It is silent although a warning is given since this warnings drowns in the set of warnings during benchmarking.

tests/test_mapper.py
```diff
@@ -43,9 +43,6 @@ def setUp(self):
         ql.set_option('clifford_premapper', 'yes')
         ql.set_option('clifford_postmapper', 'yes')
         ql.set_option('mapper', 'minextendrc')
```
jvansomeren, line 45: What about tests in test_mapper that test initial placement?

tests/test_mapper.py
```diff
@@ -43,9 +43,6 @@ def setUp(self):
...
-        ql.set_option('mapinitone2one', 'yes')
-        ql.set_option('initialplace', 'no')
-        ql.set_option('initialplace2qhorizon', '0')
```
jvansomeren, line 48: Is initial placement only enabled by adding the pass?

jvansomeren (changes requested):
See embedded comment in code.
Mostly: compatibility with old options, clarity of mapping MIP formulation to new API, feature to stop compiler on time out, EPSILON versus small float constants used in code

include/ql/ir/prim.h
pablolh: Good point, will do, in same PR but in a preliminary commit not to mess things up.

src/ql/pass/map/qubits/place_mip/detail/impl.cc
pablolh: good catch, I also removed the (now useless) cmake dependency on threads

src/ql/pass/map/qubits/place_mip/detail/impl.cc
pablolh: It is a large value that is used in place of infinity, which HiGHS doesn't seem to have a way to express.

I will extract this out-of-the-blue magic number as a named constant.

src/ql/pass/map/qubits/place_mip/detail/impl.h
pablolh: Added fail_on_timeout

src/ql/pass/map/qubits/place_mip/place_mip.cc
pablolh: The algorithm takes as input a map, mapping pairs of qubits to number of occurrence of such 2q gate in the input circuit.
What I call a gate type is a pair of qubit indices, which represents a 2q gate.
I haven't found a better designation for this, if you do please lmk.

So with the following circuit:
cnot 2, 5
cz 2, 5
cnot 2, 5

then (2, 5) will map to 3, and (2, 5) is a 2q gate "type"

src/ql/pass/map/qubits/place_mip/place_mip.cc
pablolh: Agreed, and I think it is in fact better to BY DEFAULT stop the compiler on timeout. So the new option fail_on_timeout is TRUE by default.

tests/test_mapper.py
pablolh: Indeed. I am thinking of renaming the exising mapper to "router",
As I understand, those 2 passes are independent. The IP can run without routing, although the resulting circuit will not necessarily be run-able on the topology. The router can run regardless, and always produces a runnable circuit wrt topology.
I think it makes sense to keep them separate, esp if we implement more different "strategies" for both IP and routing.

src/ql/pass/map/qubits/place_mip/detail/impl.cc
pablolh: Removed, thanks

src/ql/pass/map/qubits/map/map.cc
pablolh: I have updated the tests to not set the option anymore.

This didn't make any fail test, which is surprising. I expect the opposite, if the option was well tested and actually did something.

I will try to investigate more. But:
- the unit test I added should be IMO the new way forward to test a single pass or class. Integration tests using the Python API should be in small numbers, testing global options and overall integration. C++ unit tests are wayyy better for debugging, and they allow better code coverage and it's easier to reason at the class level, rather than at the whole program level.
- I noticed that a lot of those Python tests only test that the cqasm input compiles, they do not check that the actual output corresponds to a correct output (except those using the "golden" files), which is not strong
- There seems to be some good cleanup to do in the tests folder.

src/ql/pass/map/qubits/map/detail/mapper.h
pablolh: Fixed

tests/test_mapper.py
pablolh: Those tests don't compare the compiler output to a reference.

I do note that initial placement was previously disabled because of the license issues, and thus was not running in CI.
Those tests did not make use of initial placement before, they just output the following log line:
`InitialPlace support disabled during OpenQL build [DONE]` and IP was no-op.

That's the big issue with on/off preprocessor variables, they likely result in uncovered code in CI, except if the CI setup follows.

I can adapt test_string in mapper.cc and put the same thing in the new unit test. I've already added a similar tests that requires a log of optimization to find a perfect mapping so we can discuss relevance of this test later today.

src/ql/pass/map/qubits/place_mip/detail/impl.h
pablolh: Do you mean the old options in the mapper pass?
Or the global options

src/ql/pass/map/qubits/place_mip/detail/impl.cc
pablolh: EPSILON is small, while 1.0e5 is "infinity" (but it must not be too large, otherwise HiGHS complains... I've been exchanging emails with the dev team of HiGHS, I can ask them what exactly should be used).

src/ql/pass/map/qubits/place_mip/detail/impl.cc
pablolh: In this specific case it is worth it to add a comment indeed. This API is not the simplest... WIll do.

On the side, I just want to point out that in a lot of cases there are also drawbacks to having too many comments:
- comments and code needs to be kept in sync, and in my experience this is hard and tiresome
- comments are not needed when the code is sufficiently clear
- comments are one more (unchecked by the compiler) source of typos, errors, confusion... In the end they can make the code hard to read, even though their goal is the opposite.

In OpenQL, I disagree that there needs to be a comment for every function:
-  A function named "computeCostmax" does not need to have an accompanying comment "This function computes costmax"
- Currently there are comments for both declarations in the headers and in the implementation. Those comments are supposed to be identical I think, but it's also a lot of useless text...
I know there is some doxygen documentation but it's also fine to leave some comments out.

jvansomeren (approved):

PR #472, opened 2023-01-24
Removed compilation warnings.

rturrado:
- backend.cc: removed unused variable 'wait'.
- functions.cc/.h: changed 'expOpCnt' parameter type from Int to UInt, to avoid a signed/unsigned comparison. The same for the for loop variable 'i'.
- new_to_old.cc: removed unused variable 'for_loop'.
- old_to_new.cc: initialize 'assignable' boolean variable, to avoid its use when uninitialized.
- prim.h:
  - Warnings complained about 'data' being initialized from an uninitialized 'ncols'. This may be a false positive, due to the constructor parameter being called like the member variable. Anyway, it can be easily fixed by renaming the parameter (having parameter names equal to member names can be quite confusing).
  - made constructors receiving one single parameter explicit.
  - removed an extra semicolon at the end of the initialize function template.

### DIFF
--- a/include/ql/ir/prim.h
+++ b/include/ql/ir/prim.h
@@ -26,7 +26,7 @@ namespace prim {
  * generated tree nodes to ensure that there's no garbage in the nodes.
  */
 template <class T>
-T initialize() { return T(); };
+T initialize() { return T(); }
 
 /**
  * Serializes the given primitive object to CBOR.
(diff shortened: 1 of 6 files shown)


rturrado:
All these changes have finally gone into the main branch via  other PRs.

PR #475, opened 2023-05-01
Fix #474: assertion fires in multicore mapper

### LINKED ISSUE #474: Assert "!alters.empty()" fails in mapper in multicore setting

pablolh:
- Fix computation of is_comm_qubit in topology.cc

### DIFF
--- a/src/ql/com/tests/topology_test.cc
+++ b/src/ql/com/tests/topology_test.cc
@@ -1,0 +1,153 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include "doctest/doctest.h"
+
+#include "ql/com/topology.h"
+
+namespace ql {
+namespace com {
+
+class TopologyTest {};
+
+TEST_CASE_FIXTURE(TopologyTest, "Single core grid 2x3") {
+    std::uint64_t qubit_count = 6;
+
+/*
+
+        0 --- 1 --- 2
+        |     |     |
+        5 --- 4 --- 3
+
+*/
+
+    auto victim = ql::com::Topology(qubit_count, ql::utils::Json( R"({
+"form": "xy",
+"x_size": 2,
+"y_size": 3,
+"qubits": [
+    { "id": 0, "x": 0, "y": 0},
+    { "id": 1, "x": 0, "y": 1},
+    { "id": 2, "x": 0, "y": 2},
+    { "id": 3, "x": 1, "y": 2},
+    { "id": 4, "x": 1, "y": 1},
+    { "id": 5, "x": 1, "y": 0}
+],
+"edges": [
+    { "id": 0, "src": 0, "dst": 1},
+    { "id": 1, "src": 1, "dst": 0},
+    { "id": 2, "src": 1, "dst": 2},
+    { "id": 3, "src": 2, "dst": 1},
+    { "id": 4, "src": 2, "dst": 3},
+    { "id": 5, "src": 3, "dst": 2},
+    { "id": 6, "src": 3, "dst": 4},
+    { "id": 7, "src": 4, "dst": 3},
+    { "id": 8, "src": 4, "dst": 5},
+    { "id": 9, "src": 5, "dst": 4},
+    { "id": 10, "src": 5, "dst": 0},
+    { "id": 11, "src": 0, "dst": 5},
+    { "id": 12, "src": 1, "dst": 4},
+    { "id": 13, "src": 4, "dst": 1}
+]
+})"_json ));
+
+    CHECK_EQ(victim.get_num_cores(), 1);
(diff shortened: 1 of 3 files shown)


rturrado (commented):
Hey Pablo,

Just a couple of comments:

1) As part of my changes for issue 476, I was considering adding `num_qubits_per_core` as a member variable to topology, since it is calculated in a few places in the code. It could be just initialized at the `Topology` constructor, then used anywhere. What do you reckon? If you find it useful, you can add it in this commit. Or I can add it later in mine.

[Added_num_qubits_per_core_as_a_topology_member_helper_.patch](https://github.com/QuTech-Delft/OpenQL/files/11366443/Added_num_qubits_per_core_as_a_topology_member_helper_.patch)

2) I would add a couple of tests more for the single core case: `all qubits are communication qubits`, and `2 communication qubits for a 4-qubits core`.

Best!

pablolh:
Hi Roberto,

Originally I went for the least effort unit test that exposes the bug. In this new change it's better and I also integrated the precomputation of the number of qubits per core.

Pablo

src/ql/com/tests/topology_test.cc
```diff
@@ -0,0 +1,156 @@
...
+#include "doctest/doctest.h"
+
+#include "ql/com/topology.h"
+
+namespace ql {
+namespace com {
+
+class TopologyTest {};
+
+TEST_CASE_FIXTURE(TopologyTest, "Single core grid 2x3") {
```
pablolh, line 11: Since this class does a lot of things, including user input validation, there are still a lot of unit tests missing for e.g. checking that exceptions are thrown properly. Not in this change though.

src/ql/com/tests/topology_test.cc
```diff
@@ -0,0 +1,156 @@
...
+
+/*
+
+        0 --- 1 --- 2
+        |     |     |
+        5 --- 4 --- 3
+
+*/
+
+    // This API is weird: I don't understand why "edges" need to be specified, when "form" is "xy".
```
pablolh: @jvansomeren do you reckon that specifying a grid structure should implicitely declare the grid qubit connectivity, not just coordinates?

jvansomeren:
Hi Pablo,

Indeed the xy form just specifies that there is a coordinate system in which the qubits can be declared.
The example is the 7 qubit starmon system:

0     1
     /     \     /      \
           2        3          4
                     \     /    \      /
                        5         6

that has x-coordinates 0 to 4 and y-coordinates 0 to 2;
only every other grid point contains a qubit.

Similarly, edges are between specified pairs of qubits, not positions in the grid, and not defined by the underlying grid.

Originally there was a rectangular grid (with a qubit hat every grid point),
and a diamond grid (like the starmon 7) which is like a rectangular grid but then 45 degrees rotated and then fit in an underlying grid;
then there were odd rows (0 1) and (5 6) and even rows (2 3 4), odd columns (2), (3) and (4), and even columns (0 5) and 1 6)
but that grew too complicated.
Out of came the xy as it is now.

Best,

Hans

Op 2 mei 2023, om 10:33 heeft Pablo Le Hénaff ***@***.***> het volgende geschreven:



@pablolh commented on this pull request.

In src/ql/com/tests/topology_test.cc<https://urldefense.com/v3/__https://github.com/QuTech-Delft/OpenQL/pull/475*discussion_r1182247138__;Iw!!PAKc-5URQlI!8ugPm_sq7Nj8cs48ODZXK4UICnsXe98N3icOEYMtc5WDN46u0bRrkkvoD77T9LGDj6ZdTckFNI0N8vxhkgnkMnHOl66sv2_z$>:

> +namespace com {
+
+class TopologyTest {};
+
+TEST_CASE_FIXTURE(TopologyTest, "Single core grid 2x3") {
+    std::uint64_t qubit_count = 6;
+
+/*
+
+        0 --- 1 --- 2
+        |     |     |
+        5 --- 4 --- 3
+
+*/
+
+    // This API is weird: I don't understand why "edges" need to be specified, when "form" is "xy".


@jvansomeren<https://urldefense.com/v3/__https://github.com/jvansomeren__;!!PAKc-5URQlI!8ugPm_sq7Nj8cs48ODZXK4UICnsXe98N3icOEYMtc5WDN46u0bRrkkvoD77T9LGDj6ZdTckFNI0N8vxhkgnkMnHOlzcDpPVI$> do you reckon that specifying a grid structure should implicitely declare the grid qubit connectivity, not just coordinates?

pablolh:
Okay, then this was normal, I've removed the comment so this is ready for review

rturrado:
Beautiful, Pablo, great work!

Thanks for adding the num_qubits_per_core stuff.

Single core tests I think will be helpful.

rturrado:
Sorry, closed by mistake!

rturrado (approved):
All OK from my side.

PR #478, opened 2023-05-03
Get min hops

rturrado:
The main change of this pull request is in topology.cc, get_min_hops function:
- The `distance + 1` case is now taken into consideration.
- Doc for that function has been improved a lot. Thanks for that, Hans.

The other important change is in mapper.cc:
- I've added a `are_nearest_neighbors` function to encapsulate the `get_min_hops(source, target) == 1` check there. 
- And I've changed the two points in the file doing that check to call that function and look similar in code.
Please have a look at the function's doc and name because none of them may be appropriate.

Apart from this, other minor changes:
- added curly braces for single statement blocks in topology.cc.
- took `source == target` base case out of `connectivity == FULL` at topology.cc, get_distance.
- added a `source == target` base case in topology.cc, get_core_distance.
- added a `source == target` base case in topology.cc, get_min_hops.
- added a GetMinHops test fixture to topology_test.cc.


jvansomeren (dismissed):

src/ql/com/topology.cc
```diff
@@ -685,42 +689,73 @@ utils::UInt Topology::get_distance(Qubit source, Qubit target) const {
...
         return d;
     }
-
     return distance[source][target];
 }
 
 /**
  * Returns the distance between the given two qubits in terms of cores.
  */
 utils::UInt Topology::get_core_distance(Qubit source, Qubit target) const {
```
pablolh: This should return a bool and be renamed "are_in_same_core" or something like that

src/ql/com/topology.cc
```diff
@@ -685,42 +689,73 @@ utils::UInt Topology::get_distance(Qubit source, Qubit target) const {
...
-    QL_ASSERT(cd <= d);
-    if (cd == d) {
-        return d+2;
+    if (source == target) {
+        return 0;
+    }
+    utils::UInt distance = get_distance(source, target);
+    utils::UInt core_distance = get_core_distance(source, target);
+    QL_ASSERT(core_distance <= distance);
+    if (core_distance == distance) {
```
pablolh: at this point, distance >= 1 and core_distance is only 0 or 1
so this "if" implies `core_distance == distance == 1` and source and target being in different cores, therefore
`is_comm_qubit(source)` and `is_comm_qubit(target)` are always true, no?

I would rewrite as such, which to me is more readable:

```c++
if (!are_in_same_core(source, target) && is_comm_qubit(source) && is_comm_qubit(target)) {
  QL_ASSERT(distance == 1);
  if (num_comm_qubits == 1) {
    return 3;
  }
  return 2;
}
```
and then `core_distance` variable is useless.

Btw I think we should throw an exception when "edges" key is present or "connectivitiy" is "specified" and num_cores >= 2, in light of what Roberto told me about the assumption with multi-core connectivity.

src/ql/com/topology.cc
```diff
@@ -685,42 +689,73 @@ utils::UInt Topology::get_distance(Qubit source, Qubit target) const {
...
- * been placed yet, then at least one additional inter-core hop is needed
- * for the 2qgate, the number of hops required being at least distance+1.
+ * The minimum number of hops is always >= distance(from, to).
  *
- * We assume below that a valid path exists with distance+1 hops; this fails
- * when not all qubits in a core support connections to all other cores.
- * See the check in initialization of neighbors.
+ * Inside one core, minimum number of hops == distance.
+ *
+ * Between qubits in different cores, the minimum number of hops >= distance + 1.
```
pablolh: The distinction between "minimum number of hops" and distance is here very confusing.
A distance is a minimum number of hops already.
I think get_min_hops should be renamed with a name that has to do with routing.

src/ql/com/topology.cc
```diff
@@ -685,42 +689,73 @@ utils::UInt Topology::get_distance(Qubit source, Qubit target) const {
...
-
     return distance[source][target];
 }
 
 /**
  * Returns the distance between the given two qubits in terms of cores.
  */
 utils::UInt Topology::get_core_distance(Qubit source, Qubit target) const {
-    if (get_core_index(source) == get_core_index(target)) return 0;
+    if (source == target ||
```
pablolh: If source == target, then logically get_core_index(source) == get_core_index(target), therefore isn't that condition useless? I think previous version was OK no?

src/ql/pass/map/qubits/map/detail/mapper.cc
```diff
@@ -399,6 +399,14 @@ void Mapper::map_routed_gate(const ir::compat::GateRef &gate, Past &past) {
 
 }
 
+/**
+ * Nearest neighbors are in the same core, and at a distance of exactly one hop.
+ * This excludes the case of two communication qubits in different cores and at a distance of exactly one hop.
+ */
+utils::Bool Mapper::are_nearest_neighbors(UInt source, UInt target) const {
+    return platform->topology->get_min_hops(source, target) == 1;
```
pablolh, line 407: Looking again, get_min_hops is only used by the mapper and has to do with routing-specifics.
Maybe it's better to move get_min_hops out of topology and inside mapper. What do you think?

jvansomeren:
Yes, given the current all-to-all connectivity of multi-core.
As soon as we deviate from this, get_core_distance comes back.
Having a grid of cores was in my mind when originally writing this.
So I’m ok with rewriting this to a bool but bear in mind that it will come back, probably soon.

Op 3 mei 2023, om 12:19 heeft Pablo Le Hénaff ***@***.******@***.***>> het volgende geschreven:



@pablolh commented on this pull request.

In src/ql/com/topology.cc<https://urldefense.com/v3/__https://github.com/QuTech-Delft/OpenQL/pull/478*discussion_r1183465478__;Iw!!PAKc-5URQlI!_9CSc-bfUB6VttmJQnLa-ASjEGdPagFDt9HliDxsHEmlEF8v031McRICXy25XpkOndbx_NVEoJ40DHDcryc8ZrzEyEeyNgz3$>:

> @@ -685,42 +689,73 @@ utils::UInt Topology::get_distance(Qubit source, Qubit target) const {
         }
         return d;
     }
-
     return distance[source][target];
 }

 /**
  * Returns the distance between the given two qubits in terms of cores.
  */
 utils::UInt Topology::get_core_distance(Qubit source, Qubit target) const {


This should return a bool and be renamed "are_in_same_core" or something like that

In src/ql/com/topology.cc<https://urldefense.com/v3/__https://github.com/QuTech-Delft/OpenQL/pull/478*discussion_r1183469158__;Iw!!PAKc-5URQlI!_9CSc-bfUB6VttmJQnLa-ASjEGdPagFDt9HliDxsHEmlEF8v031McRICXy25XpkOndbx_NVEoJ40DHDcryc8ZrzEyIWyVHsn$>:

>   */
 utils::UInt Topology::get_min_hops(Qubit source, Qubit target) const {
-    utils::UInt d = get_distance(source, target);
-    utils::UInt cd = get_core_distance(source, target);
-    QL_ASSERT(cd <= d);
-    if (cd == d) {
-        return d+2;
+    if (source == target) {
+        return 0;
+    }
+    utils::UInt distance = get_distance(source, target);
+    utils::UInt core_distance = get_core_distance(source, target);
+    QL_ASSERT(core_distance <= distance);
+    if (core_distance == distance) {


at this point, distance >= 1 and core_distance is only 0 or 1
so this "if" implies core_distance == distance == 1 and source and target being in different cores, therefore
is_comm_qubit(source) and is_comm_qubit(target) are always true, no?

I would rewrite as such, which to me is more readable:

if (!are_in_same_core(source, target) && is_comm_qubit(source) && is_comm_qubit(target)) {
  QL_ASSERT(distance == 1);
  if (num_comm_qubits == 1) {
    return 3;
  } else {
    return 2;
  }
}


and then core_distance variable is useless.

Btw I think we should throw an exception when "edges" key is present or "connectivitiy" is "specified" and num_cores >= 2, in light of what Roberto told me about the assumption with multi-core connectivity.

In src/ql/com/topology.cc<https://urldefense.com/v3/__https://github.com/QuTech-Delft/OpenQL/pull/478*discussion_r1183474582__;Iw!!PAKc-5URQlI!_9CSc-bfUB6VttmJQnLa-ASjEGdPagFDt9HliDxsHEmlEF8v031McRICXy25XpkOndbx_NVEoJ40DHDcryc8ZrzEyLK2uAQt$>:

>   *
- * We assume below that a valid path exists with distance+1 hops; this fails
- * when not all qubits in a core support connections to all other cores.
- * See the check in initialization of neighbors.
+ * Inside one core, minimum number of hops == distance.
+ *
+ * Between qubits in different cores, the minimum number of hops >= distance + 1.


The distinction between "minimum number of hops" and distance is here very confusing.
A distance is a minimum number of hops already.
I think get_min_hops should be renamed with a name that has to do with routing.

jvansomeren:
The current topology avoids separating intra-core and inter-core connectivities.
Furthermore, I agree that get_min_hops knows about routing in the sense that one of hops must be able to host a 2q gate; that is weird indeed here.
But get_min_hops is also specialized now for multi-core connectivity to be all-to-all.
So this calls for splitting get_min_hops somehow to separate all this.
This splitting can only be sensibly done when we have more topology options.
So I’m fine with anything that works now, but be ready in your mind for new kinds of topologies
and new kinds of routing later.

Op 3 mei 2023, om 12:28 heeft Pablo Le Hénaff ***@***.******@***.***>> het volgende geschreven:



@pablolh commented on this pull request.

In src/ql/pass/map/qubits/map/detail/mapper.cc<https://urldefense.com/v3/__https://github.com/QuTech-Delft/OpenQL/pull/478*discussion_r1183509363__;Iw!!PAKc-5URQlI!_629NkbiEDKSm-7Ondxxe7RFy_FPo9zYfIvOiMTwQ8s5IYYXjXqkNR-mQzKNqIVgLzIiGPnLm5zLEZeGwB6Qz9cT125HjMkE$>:

> @@ -399,6 +399,14 @@ void Mapper::map_routed_gate(const ir::compat::GateRef &gate, Past &past) {

 }

+/**
+ * Nearest neighbors are in the same core, and at a distance of exactly one hop.
+ * This excludes the case of two communication qubits in different cores and at a distance of exactly one hop.
+ */
+utils::Bool Mapper::are_nearest_neighbors(UInt source, UInt target) const {
+    return platform->topology->get_min_hops(source, target) == 1;


Looking again, get_min_hops is only used by the mapper and has to do with routing-specifics.
Maybe it's better to move get_min_hops out of topology and inside mapper. What do you think?

src/ql/com/topology.cc
rturrado: Yes, good point.

src/ql/com/topology.cc
rturrado: Good. I've updated the function. I've added a comment because, despite the function's documentation, that check can be quite ilegible in some time (so any extra help to the future reader is welcome). Also, I prefer sticking to using `distance + 2` and `distance + 1` instead of `3` and `2` respectively (those `2` and `1` are also magic numbers, but more readable than `3` and `2`, IMO).

src/ql/com/topology.cc
rturrado: Well, my point here was that it saves you from calling a couple of functions. But, on the other hand, it is true that it adds an extra check, and it is an extra check that is usually not recommended, like the typical check for self assingment in an assignment operator (we shouldn't check for the unprobable case).

src/ql/pass/map/qubits/map/detail/mapper.cc
rturrado: I'm with Hans here in that there may be coming changes to Topology in a near future. I would rather keep things as they are at the moment. Even if those changes weren't coming, I would indeed separate this comment's proposal to a different commit. So, let's say, in any case, I would limit the scope of this commit to just fixing the distance cases.

src/ql/com/topology.cc
rturrado: See my comment below regarding moving `get_min_hops` out of the `Topology` class. I think this change would go together with that other change.

src/ql/com/topology.cc
rturrado: I have also removed the assert for `connectivity == FULL` because `get_core_index` already asserts that.

src/ql/com/topology.cc
rturrado: I have added a `connectivity == SPECIFIED && num_cores >= 2)` check, plus exception, at the `Topology` constructor, right after the `Handle connectivity key` section.

pablolh (dismissed):

include/ql/com/topology.h
```diff
@@ -269,6 +269,11 @@ class Topology {
      */
     utils::Bool is_comm_qubit(Qubit qubit) const;
 
+    /**
+     * Returns whether the two given qubits are communication qubits.
+     */
+    utils::Bool are_comm_qubits(Qubit source, Qubit target) const;
```
pablolh, line 275: Is this function really necessary? I think the && with two is_comm_qubits is fine.

rturrado: Not really. It is just a helper. I would keep it if it makes the code clearer. I've actually just seen another place in the code where it could also be used: `if (is_inter_core_hop(qs, qd) && (!is_comm_qubit(qs) || !is_comm_qubit(qd)))` in `get_neighbors_list` could be rewritten as `if (is_inter_core_hop(qs, qd) && !are_comm_qubit(qs, qd))`. But I leave it up to you to keep it or just use the `is_comm_qubit`.

pablolh (dismissed):

pablolh (approved):

PR #496, opened 2023-08-02
v3 without conan

rturrado:
### CMake

Updated C++ version to C++23.
Updated GitHub Actions to build in gcc/clang/msvc x Linux/MacOS/Windows x Debug/Release.
Updated dependency management to be done through FetchContent instead of git submodules.
- CMake minimum required version is now 3.25, due to FetchContent_Declare(... SYSTEM).
- lemon dependency is now fetched through a fork, which fixes C++17 & 20 compilation, and some CMake and C++ warnings.

### Reorganized tests, examples, and resources.

Reorganized tests.
- All tests are now under the `test` folder.
- All tests are `googletest`. `doctest` dependency has been removed.
- All tests are discovered. CMake code for adding tests has been removed.
- Python tests have been updated to comply to most of PEP 8 (code style guide).

Reorganized resources.
- All resources are now under the `res` folder.
- Some file names were changed.

Reorganized examples.
- All example codes are now under `example` folders within `test`.
- All example resources are now under `example` folders within `res`.

The structure of the `test` folder is:
```
  |- ql
  |  \- com, ir, pass, utils...
  \- v1x
     |- cpp
     |  \- example
     \- python
        |- example
        \- visualizer
```
The structure of the `res/v1x` folder is:
```
  |- cq
  |  |- example
  |  \- golden
  |- json
  |  \- visualizer
  \- qasm
     \- golden
```
`v1x` stands for cQasm v1.x. There may be folders such as `v3x` in the future.

### File changes

- deps: removed.
- docs:
  - Changed `from openql import openql as ql` to `import openql as ql`.
  - Changed `nqubits` to `num_qubits`.
- include/ql, source/ql:
  - Fixed warnings and compilation errors.
  - Update references to resources to point to the new location in `res`.
- python/ql.i: minor aesthetic change: split a very whole string into different lines.
- .gitmodules: removed.
- CMakeLists.txt: C++23, FetchContent, `ql` target_compile_options.
- pytest.ini: updated testpaths to point to `test/v1x/python`.


.github/workflows/test.yml
```diff
@@ -7,59 +7,140 @@ on:
...
-      run: |
-        mkdir cbuild
-        cd cbuild && cmake ../examples/cpp-standalone-example
-    - name: Build
-      run: make -C cbuild -j 5
-    - name: Test
-      run: cd tests && ../cbuild/example
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
pablolh: fetch-depth

.github/workflows/test.yml
```diff
@@ -68,109 +149,44 @@ jobs:
...
-#      # Note that conda build already does a test install in a fresh
-#      # environment, so this is a bit redundant anyway.
-#      if: matrix.os != 'windows-2016' && matrix.os != 'windows-latest'
-#      run: conda install openql --use-local
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
```
pablolh, line 154: you're removing testing for 4 versions of Python
what's the reasoning?
Personally I'm fine with that, but then do we no longer support Python < 3.11?

.github/workflows/test.yml
pablolh: Also think of updating assets.yml

README.md
```diff
@@ -17,4 +17,4 @@ can be simulated on the QX simulator.
 OpenQL's source code is released under the Apache 2.0 license.
 
 For detailed user and contributor documentation, visit the
```
pablolh, line 19: just noticed the extra "the"

setup.py
```diff
@@ -192,7 +192,6 @@ def run(self):
                     cmd = cmd['--parallel'][nprocs]
                 elif not sys.platform.startswith('win'):
                     cmd = cmd['--']['-j'][nprocs]
-            cmd & FG
```
pablolh, line 195: Is that why CMake was running twice every time I was building the Python package? Good fix!

src/ql/ir/compat/detail/cqasm_reader.h
```diff
@@ -19,12 +19,12 @@ namespace detail {
 
 // Shorthands for namespaces.
 namespace lqt { using namespace ::cqasm::tree; }
-namespace lqs { using namespace ::cqasm::semantic; }
-namespace lqv { using namespace ::cqasm::values; }
-namespace lqtyp { using namespace ::cqasm::types; }
-namespace lqi { using namespace ::cqasm::instruction; }
-namespace lqa { using namespace ::cqasm::analyzer; }
-namespace lqp { using namespace ::cqasm::parser; }
+namespace lqs { using namespace ::cqasm ::v1x::semantic; }
```
pablolh: spurious space?

src/ql/utils/vcd.cc
```diff
@@ -83,7 +83,7 @@ void Vcd::change(Int var, Int timestamp, const Str &value)
 }
 
 
-void Vcd::change(Int var, Int timestamp, Int value) {
```
pablolh, line 86: I understand that the variable was not used, however here the variable names are clearly useful to know what the intent of the code was... or would have been...
This class is not inherited, so here is the only way to know what the code was...

This code has no owner anymore unfortunately. I would suggest not touching it.

test/ql/com/ddg/ddg.cc
```diff
@@ -1,13 +1,16 @@
-#include "ql/ir/compat/compat.h"
-#include "ql/ir/old_to_new.h"
 #include "ql/com/ddg/build.h"
-#include "ql/com/ddg/ops.h"
```
pablolh, line 4: I think the intent of placing tests in src was to distinguish between
- unit tests, or so-called unit tests, accompanying the source
- integration tests, in tests/ folder
which is a pretty reasonable organization to me. Unfrotunately in OpenQL it's not well done, there are a bit of integration tests everywhere and almost no unit test, which is very annoying to know the source of an issue and the intent of the code.

Normally the unit tests link against only the piece they test.
OpenQL is just a massive CMake library so it doesn't bring any benefit.
Unfortunately my CMake change didn't go through, see https://github.com/QuTech-Delft/OpenQL/pull/460 where I tried to split CMake targets.

pablolh: If we had smaller CMake targets, having multiple test executables wouldn't be an issue, I think.

test/v1x/cpp/example/cpp_standalone.cc
```diff
@@ -0,0 +1,31 @@
+#include <openql>
+
+#include <iostream>
+
+
+int main(int argc, char **argv) {
```
pablolh: I think this could be a google test as the others.
It's an example but can totally be a test in the big test executable.

pablolh: Will also save on link time

pablolh:
few comments I think this can be merged quickly!

.github/workflows/test.yml
rturrado: Done! Well spotted!

I haven't tested `assets.yml` for none of my last changes in libqasm, OpenQL, and QX-simulator. I guess we'll address possible issues there when we have to cross those bridges.

.github/workflows/test.yml
rturrado: Yes, sorry about this. We can put them back. I used the QX-simulator's `test.yml` as a template, and I saw that only version `3.11` was used there.

README.md
rturrado: Haha, that was a hidden gem! 😄

setup.py
rturrado: Wow. I thought clang-tidy had spotted that line as "never executing", but now I see the code, I don't see why it wouldn't be executing.

src/ql/ir/compat/detail/cqasm_reader.h
rturrado: Wow! I wonder how that code was working! 😮

src/ql/utils/vcd.cc
rturrado: This is a good comment. I should have, in general, just commented out the variable names.

I've put the variable names back in again for this particular case. They're commented out though to avoid compilation warnings/errors.

test/v1x/cpp/example/cpp_standalone.cc
rturrado: My idea was to leave it as an example. I think it wasn't built at the moment. But I see no reason either to add it as a test, yes.

test/ql/com/ddg/ddg.cc
rturrado: I'll have to dedicate some time to go through the comments in that other PR! 😄 

We could perfectly try to not mix unit and integration tests in the same binary. That's perfectly sensible. Having, for example, an `OpenQL_unit_test` and an `OpenQL_integration_test`.

What I'm more in favour of, maybe because I think it's more an standard way, is to separate tests from source code. But, again, I may be very wrong here.

I'll leave this conversation unresolved so that we can talk about it next week.

.github/workflows/test.yml
pablolh: I think it's fair

test/ql/com/ddg/ddg.cc
pablolh: I think with CMake it's indeed standard to separate source and tests, fair point.
Let's not discuss the tests problem too much in this PR anyway - it's devoted to sth else.

I'm OK with them moving, actually your tests/CMakeLists.txt is a lot cleaner than the previous testing section in the CMakeLists at the root

pablolh (approved):

PR #497, opened 2023-09-08
Toy v1 OpenQL v1/v3 reader

rturrado:
OpenQL update to work with the latest libqasm.

The latest libqasm:
- adds ANTLR for parsing cQASM 3.0 files, and
- builds only via Conan.

This commit brings the following changes to OpenQL:
- Can only be built via Conan.
- Works with the latest libqasm.
- Changes the Reader's `read` function to enroute parsing of cQASM 3.0 files to the libqasm's v3 parser (ANTLR).
- Fixes all the GitHub Actions builds.


pablolh (commented):
Nice!
I think it might need some modifications to the docs/ folder and the README.md because they probably still refer to CMake and submodules?

rturrado:
> Nice! I think it might need some modifications to the docs/ folder and the README.md because they probably still refer to CMake and submodules?

Hey Pablo! Yes, very good point! I definitely have to revise the README and the documentation in general.

pablolh (approved):

PR #501, opened 2023-09-27
Add conan profiles

rturrado:
Welcome `conan build . -pr=<path/to/profile> -b missing`! 😊

This PR tries to address issue [500](https://github.com/QuTech-Delft/OpenQL/issues/500).

- Added a group of predefined profiles to `conan/profiles`.
  These profiles cover the most common build scenarios: tests on/off * debug/release * unitary decomposition on/off.
- Updated README.md.
- Updated `test.yml` to use conan profiles.


.github/workflows/test.yml
```diff
@@ -43,7 +43,10 @@ jobs:
       - name: Configure and build
         run: |
           conan profile detect
-          conan build . -s:h compiler.cppstd=23 -s:h openql/*:build_type=${{ matrix.build_type }} -o openql/*:build_tests=True -o openql/*:disable_unitary=True -b missing
+          build_type=${{ matrix.build_type }}
+          lowercase_build_type=$(echo ${build_type} | tr '[:upper:]' '[:lower:]')
```
pablolh, line 47: I'm wondering like in the previous PR, if we can set `matrix.build_type` to the lowercase version and ask CMake to build in the lowercase build dir, so that this transformation is not needed.

pablolh (dismissed):

.github/workflows/test.yml
rturrado: Thanks for the comment Pablo. I valued that option together with changing the profile names to start with uppercase, but I finally decided I like both 1) build directories starting with uppercase, and 2) lowercase for the profile names.

BTW, testing that line took me a while. It seems that `tr` is POSIX standard, but things I tried before, like `${build_type@L}` or `${build_type,,}` were not. 😊

pablolh (dismissed):

pablolh (approved):

PR #513, opened 2024-01-09
Release 0.12.2

rturrado:
Fix `upload-release-asset` steps in `.github/workflow/assets.yml` for `windows-x64` job.

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -206,7 +206,6 @@ jobs:
         run: |
           echo "WHEEL_NAME=$(Get-ChildItem -name *.whl)" >> $env:GITHUB_OUTPUT
           echo "WHEEL_NAME=$(Get-ChildItem -name *.whl)" >> $env:GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/assets.yml
```diff
@@ -162,8 +162,8 @@ jobs:
         id: wheel
         working-directory: wheelhouse
         run: |
-          echo "WHEEL_NAME=$(ls *.whl)" >> "$GITHUB_OUTPUT"
```
pablolh, line 165: why remove the quotes now?

rturrado: Because that's something I had tried, to see if it was the source of the problem, but it wasn't. So I've preferred to leave them as they were before, since I think the code looks clearer.

pablolh (approved):

.github/workflows/assets.yml
pablolh: I heard it's actually a pretty good practice to quote variables in bash
Otherwise it might behave not like you expect when the variable is a string containing separators
But here we're safe

